PR #12, opened 2018-12-11
[Open-AIS] Pull request (fix validation errors)

kuenzge:
Change Log
- We did a full revision (manually line by line) to adapt for the 1.1 specifications best
- the bit-flag-fields were reverted from opaque to Unsigned Integer with appropriate length
- Unsigned Integer has been applied instead of Integer as type wherever appropriate.
- The Core-Object-Links were reverted from String to Corelnk
- We changed the length definition of strings back to <RangeEnumeration>0-16 bytes</RangeEnumeration> to support the shared resource definitions best, and to keep some uniformity in the definitions.
- We used the - for the range definitions with the exception of defining ranges that run from negative to positive, where we used <RangeEnumeration>-128..127</RangeEnumeration> to ease readability.
- We systematically removed residual issues that were connected with our last-minute-changes to W operation instead of using E operation with a payload as event triggering method as originally designed. (still the optimal design would be an event trigger (e.g. scene recall, color set etc.) that allows a payload together with the execution.

### DIFF
--- a/3387.xml
+++ b/3387.xml
@@ -19,7 +19,7 @@ ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEAL
 		<Description1><![CDATA[The 'oA Basic Control' represents a common basic interface for a Control Function. This allows to have vendor specific (executing) control functions using diverse object numbers to be interfaced for the basic functionality in a common way. The object provides the basic functionality required for an OpenAIS control object.]]></Description1>
 		<ObjectID>3387</ObjectID>
 		<ObjectURN>urn:oma:lwm2m:ext:3387</ObjectURN>
-		<LWM2MVersion>1.0</LWM2MVersion>
+		<LWM2MVersion>1.1</LWM2MVersion>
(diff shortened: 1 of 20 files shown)


jpradocueva:
@kuenzge OMA perform two validations: the schema validation or xml and the TS validation.
The first xml validation has failed due to the fact that the ObjectVersion has been upgraded to 1.1 but the schema is still to point to v1.0.
If you want these objects to be compatible with version 1.1 of the LwM2M protocol, then you need to change the schema to LWM2M-v1_1.xsd. I pasting below the line that contains the correct schema, LwM2M-v1_1.xsd. 

```<LWM2M xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="http://www.openmobilealliance.org/tech/profiles/LWM2M-v1_1.xsd">```

```
Note: You don't need to change the whole line just the schema file version to (-v1_1): "LWM2M-v1_1.xsd"
After this change I will perform both validations.
```

jpradocueva:
This PR was superseded by PR #13.

PR #32, opened 2019-01-02
[TUT] ObjID 10300 - LwM2M Meta Object

jpradocueva:
Submitted by OMNA staff on behalf of Tampere University of Technology.

new pull request template

### DIFF
--- a/10300.xml
+++ b/10300.xml
@@ -1,0 +1,83 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!--Copyright 2018 Tampere University of Technology. 
+
+Redistribution and use in source and binary forms, with or without modification, are permitted provided that the following conditions are met:
+
+1. Redistributions of source code must retain the above copyright notice, this list of conditions and the following disclaimer. 
+
+2. Redistributions in binary form must reproduce the above copyright notice, this list of conditions and the following disclaimer in the documentation and/or other materials provided with the distribution.
+
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.-->
+<LWM2M xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="http://www.openmobilealliance.org/tech/profiles/LWM2M.xsd">
+  <Object ObjectType="MODefinition">
+    <Name>LWM2M Meta Object</Name>
+    <Description1><![CDATA[Use this meta object to provide one or more locations for XML files defining vendor-specific objects. ]]></Description1>
+    <ObjectID>10300</ObjectID>
+    <ObjectURN>urn:oma:lwm2m:x:10300</ObjectURN>
+    <LWM2MVersion>1.0</LWM2MVersion>
+    <ObjectVersion>1.0</ObjectVersion>
+    <MultipleInstances>Multiple</MultipleInstances>
+    <Mandatory>Optional</Mandatory>
+    <Resources>
+      <Item ID="0">
+        <Name>ObjectID</Name>
+        <Operations>R</Operations>
+        <MultipleInstances>Single</MultipleInstances>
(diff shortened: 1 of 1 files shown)


10300.xml
```diff
@@ -0,0 +1,83 @@
...
+
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.-->
+<LWM2M xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="http://www.openmobilealliance.org/tech/profiles/LWM2M.xsd">
+  <Object ObjectType="MODefinition">
+    <Name>LWM2M Meta Object</Name>
+    <Description1><![CDATA[Use this meta object to provide one or more locations for XML files defining vendor-specific objects. ]]></Description1>
```
akeranen: I was wondering what is the use case here if XML files are always available via OMNA registry?

10300.xml
```diff
@@ -0,0 +1,83 @@
...
+        <Name>MD5</Name>
+        <Operations>R</Operations>
+        <MultipleInstances>Single</MultipleInstances>
+        <Mandatory>Mandatory</Mandatory>
+        <Type>String</Type>
+        <RangeEnumeration>
+        </RangeEnumeration>
+        <Units>
+        </Units>
+        <Description><![CDATA[MD5 value of the XML file in the URL(s)]]></Description>
```
akeranen: Would more modern algorithm make sense since MD5 is broken as crypto hash? Should there be crypto agility, e.g., with another resource to define what is the algorithm used, and a default algo if that is missing?

10300.xml
```diff
@@ -0,0 +1,83 @@
...
+      </Item>
+      <Item ID="3">
+        <Name>URL</Name>
+        <Operations>R</Operations>
+        <MultipleInstances>Multiple</MultipleInstances>
+        <Mandatory>Mandatory</Mandatory>
+        <Type>String</Type>
```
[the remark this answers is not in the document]
bsilverajan: I think this would be a really good idea if both ObjectID and ObjectURN can be defined as reusable LWM2M resources, for example Resource ID 4002 and 4003.

10300.xml
bsilverajan: XML definitions are available in some cases. In others, such as x-label objects produced by individuals and companies in the 26241 - 32768 range, as well as x-label bulk objects, the XML definitions are not available in the registry. 

For these kinds of objects, the LWM2M Meta Object allows the XML definitions to be hosted and discovered from 3rd party services, and possibly to be retrieved and verified with M2M communication.

10300.xml
bsilverajan: This is a good idea, and it was discussed earlier in #406 [https://github.com/OpenMobileAlliance/OMA_LwM2M_for_Developers/issues/406](url)

I have redefined the Meta Object and removed the MD5 field and instead introducing 2 fields in its place: an SHAType and a ChecksumValue:

	<Item ID="3">
        <Name>SHAType</Name>
        <Operations>R</Operations>
        <MultipleInstances>Single</MultipleInstances>
        <Mandatory>Optional</Mandatory>
        <Type>Integer</Type>
        <RangeEnumeration>0-6</RangeEnumeration>
        <Units></Units>
        <Description><![CDATA[The Secure Hash Algorithm used to compute the message digest for the XML file in the URL(s). Currently defined values are "0" (no checksum used), "1" (for MD5), "2" (for SHA1), "3" (for SHA224), "4" (for SHA256), "5" (for SHA384) and "6" (for SHA512). ]]></Description>
      </Item>

	<Item ID="4">
		<Name>ChecksumValue</Name>
		<Operations>R</Operations>
		<MultipleInstances>Single</MultipleInstances>
		<Mandatory>Optional</Mandatory>
		<Type>String</Type>
		<RangeEnumeration></RangeEnumeration>
		<Units></Units>
		<Description><![CDATA[If the SHAType value is "0" then this string must be ignored. If the SHAType value is 1-6, this field contains a hexadecimal encoded checksum value based on the SHA function used. Valid characters for the string are: 0123456789abcdef.  The length of the string must correspond exactly to the respective digest. For example, a file "foo" containing "bar"  as its content would have the following string values: For SHAType 1 (MD5), 128 bits in 32 characters:  c157a79031e1c40f85931829bc5fc552  For SHAType 2 (SHA1), 160 bits in 40 characters:  e242ed3bffccdf271b7fbaf34ed72d089537b42f]]></Description>
			</Item>


@jpradocueva Should I resubmit this as a new registration based on your advice in #406 [https://github.com/OpenMobileAlliance/OMA_LwM2M_for_Developers/issues/406](url) also because of a change in the Organization name in 2019 (Tampere University of Technology is now Tampere University)? If so, should I resubmit this with the same ObjID 10300?

10300.xml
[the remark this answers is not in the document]
jpradocueva: The concept of reusable is applied only to Resources. It is possible to register those separately from the Object. The URN is unique to each Object.

10300.xml
jpradocueva: @bsilverajan I suggest to continue using this PR, there is no need fo resubmit it now that a specific branch and PR has been created.
Feel free to update this PR with the changes that you suggest, we will change the name in the registry file to Tampere University, TU.

jpradocueva:
This PR has been superseded by PR #66, therefore it is closed here.

PR #33, opened 2019-01-02
[HuaXiang] Device Data Container - 10309

jpradocueva:
Submitted on behalf of the submitter

new pull request template

### DIFF
--- a/10309.xml
+++ b/10309.xml
@@ -7,7 +7,7 @@
     <Name>DeviceDataContainer</Name>
     <Description1><![CDATA[DeviceDataContainer Object of HuaXiang Company.]]></Description1>
     <ObjectID>10309</ObjectID>
-    <ObjectURN>oma:lwm2m:x:10309</ObjectURN>
+    <ObjectURN>urn:oma:lwm2m:x:10309</ObjectURN>
     <LWM2MVersion>1.1</LWM2MVersion>
     <ObjectVersion>1.0</ObjectVersion>
     <MultipleInstances>Multiple</MultipleInstances>
--- a/10309.xml
+++ b/10309.xml
@@ -6,8 +6,8 @@
   <Object ObjectType="MODefinition">
     <Name>DeviceDataContainer</Name>
     <Description1><![CDATA[DeviceDataContainer Object of HuaXiang Company.]]></Description1>
-    <ObjectID>22222</ObjectID>
-    <ObjectURN>oma:lwm2m:x:22222:1.0</ObjectURN>
+    <ObjectID>10309</ObjectID>
+    <ObjectURN>oma:lwm2m:x:10309</ObjectURN>
     <LWM2MVersion>1.1</LWM2MVersion>
     <ObjectVersion>1.0</ObjectVersion>
     <MultipleInstances>Multiple</MultipleInstances>
--- a/10309.xml
+++ b/10309.xml
@@ -1,0 +1,74 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!--Copyright 2018 Hua Xiang IOT. All rights reserved.
+
+-->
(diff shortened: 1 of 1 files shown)


10309.xml
```diff
@@ -0,0 +1,74 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!--Copyright 2018 Hua Xiang IOT. All rights reserved.
+
+-->
+<LWM2M xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="http://www.openmobilealliance.org/tech/profiles/LWM2M-v1_1.xsd">
+  <Object ObjectType="MODefinition">
+    <Name>DeviceDataContainer</Name>
```
akeranen: The name is quite generic for a relatively specific object

10309.xml
```diff
@@ -0,0 +1,74 @@
...
+      <Item ID="1">
+        <Name>BToothstate</Name>
+        <Operations>RW</Operations>
+        <MultipleInstances>Multiple</MultipleInstances>
+        <Mandatory>Mandatory</Mandatory>
+        <Type>Integer</Type>
+        <RangeEnumeration>0-1</RangeEnumeration>
+        <Units>
+        </Units>
+        <Description><![CDATA[BlueTooth State]]></Description>
```
akeranen: More details on description would be useful. Is it on/off state? Maybe one of the reusable resources would be good here? Same comment applies to other state resources below.

10309.xml
```diff
@@ -0,0 +1,74 @@
...
+    <ObjectVersion>1.0</ObjectVersion>
+    <MultipleInstances>Multiple</MultipleInstances>
+    <Mandatory>Optional</Mandatory>
+    <Resources>
+      <Item ID="1">
+        <Name>BToothstate</Name>
+        <Operations>RW</Operations>
+        <MultipleInstances>Multiple</MultipleInstances>
+        <Mandatory>Mandatory</Mandatory>
+        <Type>Integer</Type>
```
akeranen: Maybe boolean value is better here? Same comment applies to other state resources below

10309.xml
```diff
@@ -0,0 +1,74 @@
...
+        <RangeEnumeration>0-1</RangeEnumeration>
+        <Units>
+        </Units>
+        <Description><![CDATA[Light State]]></Description>
+      </Item>
+      <Item ID="5">
+        <Name>UpTime</Name>
```
akeranen: Would on-time re-usable resource (5852) fit here?

10309.xml
```diff
@@ -0,0 +1,74 @@
...
+        <Operations>RW</Operations>
+        <MultipleInstances>Multiple</MultipleInstances>
+        <Mandatory>Mandatory</Mandatory>
+        <Type>String</Type>
```
akeranen: Probably numeric type is better

10309.xml
JiaDeng-Git: We wanted to keep some information about the device, so I blurred the name.We're sorry about that.We will uphold an open and cooperative attitude to further improve the object.
Therefore, we change the name of the object to ShareParkingLock.
The lock is used to control each parking space and upload the lock information via wireless network.

10309.xml
JiaDeng-Git: Yes,it's on/off state.
In the modified version, we deleted the resource.

10309.xml
JiaDeng-Git: Yes, your suggestion is pretty good.
We want this time down to seconds.

10309.xml
JiaDeng-Git: We add a little bit of notation to distinguish between the different times.However, after thinking about it, we also think that we should change to numeric.

JiaDeng-Git:
We have submitted a new object, please review it.

10309.xml
JiaDeng-Git: That's true,boolean value is better.

jpradocueva:
This PR has been superseded by PR#76

PR #66, opened 2019-01-25
[TUT] LwM2M Meta Object - ObjID 10300.xml

bsilverajan:
This pull request updates PR #32 with a new Meta Object definition.

Changelog: 

* Updated "Tampere University of Technology" to "Tampere University"
* Removed "MD5" resource and added "SHAType" and "ChecksumValue" resources

### DIFF
--- a/10300.xml
+++ b/10300.xml
@@ -1,6 +1,8 @@
 <?xml version="1.0" encoding="utf-8"?>
 
-<!--Copyright 2019 Tampere University. 
+<!--
+
+Copyright 2019 Tampere University. 
 
 Redistribution and use in source and binary forms, with or without modification, are permitted provided that the following conditions are met:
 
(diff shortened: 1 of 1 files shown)


10300.xml
```diff
@@ -0,0 +1,87 @@
...
+      </Item>
+      <Item ID="4000">
+        <Name>ObjectInstanceHandle</Name>
+        <Operations>R</Operations>
+        <MultipleInstances>Single</MultipleInstances>
+        <Mandatory>Optional</Mandatory>
+        <Type>Objlnk</Type>
+        <RangeEnumeration></RangeEnumeration>
+        <Units></Units>
+        <Description><![CDATA[Reusable LWM2M resource. This can be used to differentiate object instances having the same object ID and URN (for example, for Object IDs in the range 26241 - 32768)]]></Description>
```
akeranen: Unfortunately the description of a re-usable resource can't be changed but currently "how to use the re-usable resource" should be described in the object description. Long term, we should fix this in the schema.

10300.xml
```diff
@@ -0,0 +1,87 @@
...
+      </Item>
+      <Item ID="2">
+        <Name>URL</Name>
+        <Operations>R</Operations>
+        <MultipleInstances>Multiple</MultipleInstances>
+        <Mandatory>Mandatory</Mandatory>
+        <Type>String</Type>
+        <RangeEnumeration></RangeEnumeration>
+        <Units></Units>
+        <Description><![CDATA[Location of the XML file containing the object definition of the vendor-specfic object. Multiple instances can be used if several alternative locations are supplied.]]></Description>
```
akeranen: As discussed in the earlier PR, this could be turned into re-usable resource

10300.xml
```diff
@@ -0,0 +1,87 @@
...
+      </Item>
+	  <Item ID="3">
+        <Name>SHAType</Name>
+        <Operations>R</Operations>
+        <MultipleInstances>Single</MultipleInstances>
+        <Mandatory>Optional</Mandatory>
+        <Type>Integer</Type>
+        <RangeEnumeration>0-6</RangeEnumeration>
+        <Units></Units>
+        <Description><![CDATA[The Secure Hash Algorithm used to compute the message digest for the XML file in the URL(s). Currently defined values are "0" (no checksum used), "1" (for MD5), "2" (for SHA1), "3" (for SHA224), "4" (for SHA256), "5" (for SHA384) and "6" (for SHA512). ]]></Description>
```
akeranen: MD5, and even SHA-1, aren't secure anymore

10300.xml
```diff
@@ -0,0 +1,87 @@
...
+      </Item>
+	  <Item ID="4">
+				<Name>ChecksumValue</Name>
+				<Operations>R</Operations>
+				<MultipleInstances>Single</MultipleInstances>
+				<Mandatory>Optional</Mandatory>
+				<Type>String</Type>
+				<RangeEnumeration></RangeEnumeration>
+				<Units></Units>
+				<Description><![CDATA[If the SHAType value is "0" then this string must be ignored. If the SHAType value is 1-6, this field contains a hexadecimal encoded checksum value based on the SHA function used. Valid characters for the string are: 0123456789abcdef.  The length of the string must correspond exactly to the respective digest. For example, a file "foo" containing "bar"  as its content would have the following string values: For SHAType 1 (MD5), 128 bits in 32 characters:  c157a79031e1c40f85931829bc5fc552. For SHAType 2 (SHA1), 160 bits in 40 characters: e242ed3bffccdf271b7fbaf34ed72d089537b42f]]></Description>
```
akeranen: This should probably be described in more detail to avoid interop issues. I would have assumed only the contents of the file go through the hash, but the text seems to imply the file name matters too?

akeranen (changes requested):

10300.xml
bsilverajan: I'm not entirely sure exactly what should be changed here, as this particular resource is the ObjectInstanceHandle resource, defined in the OMNA registry as a reusable resource with the Resource ID 4000.

10300.xml
bsilverajan: That would be good to define. I'd also recommend that in addition to the ObjectInstanceHandle that LWM2M already defines as a reusable resource, and the URL that you are proposing to turn into a reusable resource, the ObjectID and the ObjectURN can also be defined as reusable resources.

10300.xml
bsilverajan: I opted to retain MD5 and SHA-1 for backwards compatibility, but offering only SHA2 and SHA3 is a good idea.

10300.xml
bsilverajan: Sorry for the confusing text. The file name has no bearing on the checksum value, only the contents

10300.xml
akeranen: The description should match the description in OMNA

akeranen: The description should match the description in OMNA

10300.xml
akeranen: This was discussed in the IPSO call today and a third option was suggested: we could potentially make a new data type for URLs that could be used instead of String. Would that be good here? However, that could come earliest with next version of the LwM2M spec.

bsilverajan: Certainly if a new URI datatype exists, that can be used. It sounds as if that is an option and not a certainty for the next LWM2M spec version. Can this object now be registered with a String type for URLs and URNs and updated in future for using new datatypes?

10300.xml
bsilverajan: As it is written now, the resource definition conforms to the current practice for the OMNA registry. 
Do note that the registry indicates that for reusable resources, the value of the Description attribute can be changed. Only those attributes marked with an asterisk contain immutable values and Description is not one of them.

10300.xml
bsilverajan: To clarify, this is an optional field with the original intent being to provide a simple means to verify that the xml definition file indeed contains the correct object description. However, I do see and agree that it is a good idea to consider stronger protection for file integrity and malicious/unintentional tampering. So, let's drop MD5 and SHA-1 and go with SHA2 and SHA3. Any additional computational or storage complexity is in fact very minimal.

ScottTaftPotter:
@bsilverajan is to update resource 3 to remove MD5 and SHA1 from hash types. resource 2 to remain string for now, change to new type coming in version 1.2.

akeranen:
And reference the URI RFC for syntax of the URI string

bsilverajan:
@ScottTaftPotter @akeranen Thanks for all the feedback. I have created a new object definition based on your comments and the updated file appears available as a second commit in this branch. Notably, the description of reusable resource ObjectInstanceHandle would be removed, MD5 and SHA1 algorithms would no longer be supported and both the ObjectURN and URI resources would reference the respective RFCs.

jpradocueva:
In the latest PR the DDF.xml and Object 10300.xml validates correctly.
**The Object/PR is ready for the Working Group validation**

PR #94, opened 2019-04-02
Fix broken xml in 10299.xml

lhotari:
Schema validation failed for this file. There was a typo in the file.


jpradocueva (approved):
10299 has passed validation.
It appears an error on the submitter DDF. Further investigation is needed by the staff.

PR #112, opened 2019-04-12
[ADMIN] Vorto links removed

seanmcilroy29:
Vorto links removed from the DDF file


jpradocueva (approved):
The changes also validate correctly.

Issue #140, opened 2019-04-29
[South East Water] Gas Readings - 10313


github-oma:
<!DOCTYPE html><html><head><meta charset="utf-8" /></head><body><table><tr><td>New OMNA registration request</td><td>2019-04-29 06:27:33 UTC</td></tr><tr><td>Submitter's name</td><td>Andrew Forster-Knight</td></tr><tr><td>Defined By</td><td>Private company</td></tr><tr><td>Organization</td><td>South East Water</td></tr><tr><td>Type of registration</td><td>Object</td></tr><tr><td>Object Name</td><td>Gas Readings</td></tr><tr><td>Object Description</td><td>Object to periodically log readings from any gas sensor (eg H2S, VOCx, methane etc)</td></tr><tr><td>Download Object</td><td><a href="https://devtoolkit.openmobilealliance.org/OEditor/MODownloadZip?omna=6665c2b1-193d-4411-a2c7-b6cfad9aa673_Object.xml">Link to object (zip)</a></td></tr><tr><td></td><td><a href="https://devtoolkit.openmobilealliance.org/OEditor/OMNA/6665c2b1-193d-4411-a2c7-b6cfad9aa673_Object.xml">Link to object</a></td></tr><tr><td>Load the object into the tool</td><td><a href="https://devtoolkit.openmobilealliance.org/OEditor/LWMOView?omna=6665c2b1-193d-4411-a2c7-b6cfad9aa673_Object.xml">Load Editor with object</a></td></tr>

seanmcilroy29:
Thank you for your registration.

- [ ] The object ID allocation for this object will be reserved as 10313
- [ ] The object URN allocation for this object will be reserved as oma:lwm2m❌10313

A new branch South East Water has been created in the LwM2M registry for the WG to review your object

 - [ ] Please submit a Pull Request with the content of your object against the branch: https://github.com/OpenMobileAlliance/lwm2m-registry/tree/SE-Water

**Observations to be addressed**
  - [ ] Object version updated to 1.0
  - [ ] LwM2M verision updated 1.0
  - [ ] Object ID updated to 10313
  - [ ] Object urn updated to urn:oma:lwm2m❌10313
  - [ ] The description in your submission is ```Object to periodically log readings from any gas sensor (eg H2S, VOCx, methane etc)``` but in the object the description is ```Periodic gas sensor measurements```

Steps | Actions
-- | --
1 | Submit object using the tool
[...]

PR #141, opened 2019-04-30
Object 10313 Reserved

seanmcilroy29:
Object 10313 Reserved - see issue #140


jpradocueva:
@seanmcilroy29 check the value of the URN, it doesn't match the ObjectID.

seanmcilroy29:
@jpradocueva - Yes i saw and have updated the urn

Issue #139, opened 2019-04-29
[South East Water] Reusable resource Sensor Warm-up time - 6041


github-oma:
<!DOCTYPE html><html><head><meta charset="utf-8" /></head><body><table><tr><td>New OMNA registration request</td><td>2019-04-29 06:24:11 UTC</td></tr><tr><td>Submitter's name</td><td>Andrew Forster-Knight</td></tr><tr><td>Defined By</td><td>Private company</td></tr><tr><td>Organization</td><td>South East Water</td></tr><tr><td>Type of registration</td><td>Reusable resources</td></tr><tr><td>Download Object</td><td><a href="https://devtoolkit.openmobilealliance.org/OEditor/MODownloadZip?omna=4318c9dc-90d8-4a5c-a40f-214db4c822ea_Object.xml">Link to object (zip)</a></td></tr><tr><td></td><td><a href="https://devtoolkit.openmobilealliance.org/OEditor/OMNA/4318c9dc-90d8-4a5c-a40f-214db4c822ea_Object.xml">Link to object</a></td></tr><tr><td>Load the object into the tool</td><td><a href="https://devtoolkit.openmobilealliance.org/OEditor/LWMOView?omna=4318c9dc-90d8-4a5c-a40f-214db4c822ea_Object.xml">Load Editor with object</a></td></tr>

seanmcilroy29:
See PR #146

jpradocueva:
See PR #167 
Previous PRs are superseded by #167

seanmcilroy29:
Resource Agreed See - #167 #243

PR #146, opened 2019-04-30
[SE Water] RR - 6041 Sensor Warm-up time

seanmcilroy29:
South East Water - RR - 6041 Sensor Warm-up time
See Issue - #139

new pull request template


jpradocueva:
This PR is closed and its content moved to PR #149.
In order to perform the validation of an Object the Object and new Reusable Resources need to be in the same PR.

PR #188, opened 2019-07-02
Editorial fixes to Common.xml

akeranen:


Common.xml
```diff
@@ -18,7 +18,7 @@
         <Submitter>OMA</Submitter>
         <!-- Description of the reusable resource -->
         <Description>
-          <![CDATA[The object link is used to refer an Instance of a given Object. An Object link value is composed of two concatenated 16-bits unsigned integers following the Network Byte Order convention. The Most Significant Halfword is an ObjectID, the Least Significant Hafword is an ObjectInstance ID.An Object Link referencing no Object Instance will contain the concatenation of 2 MAX-ID values (null link. Test 5]]>
+          <![CDATA[The object link is used to refer an Instance of a given Object. An Object link value is composed of two concatenated 16-bits unsigned integers following the Network Byte Order convention. The Most Significant Halfword is an ObjectID, the Least Significant Hafword is an ObjectInstance ID. An Object Link referencing no Object Instance will contain the concatenation of 2 MAX-ID values.]>
```
bsilverajan: A typo remains in this line: "the Least Significant Hafword", which should also be rectified

akeranen: Thanks Bill! Fixed now.

jpradocueva:
@akeranen if this Pull Request is accepted, then we will have a large number of validation failures.
According to the LwM2M TS - when reusing a Reusable Resource - the description of the Reusable Resource in the new Object cannot be changed.

By changing the description, even if it is just a minor editorial change, the validator tool will raise a validation failure in a large number of objects already registered due to a mismatch in the description.
The way to resolve these validations errors is by changing the description of the affected Objects.
**Do you want us to quantified how many Object we will have to change?**

akeranen:
@jpradocueva yes, we should do editorial update on all the affected objects and update version numbers accordingly. But we should batch this change with any other editorial changes we want to make. No need to quantify the impact for now.

jpradocueva:
@akeranen minor changes in the description should not trigger an update in the version of the Object, but it will be good to have a better guidelines of what trigger a change on the Object version.

akeranen:
Superseded by #192 and #193

PR #193, opened 2019-07-11
Full stops for all RR descriptions

akeranen:
Unified use of full stops for RR descriptions

### DIFF
--- a/Common.xml
+++ b/Common.xml
@@ -246,7 +246,7 @@
         
         <Submitter>IPSO</Submitter>
 
-        <Description><![CDATA[The polarity of a digital input as a Boolean (0 = Normal, 1= Reversed)]]></Description>
+        <Description><![CDATA[The polarity of a digital input as a Boolean (0 = Normal, 1= Reversed).]]></Description>
         
         <TS></TS>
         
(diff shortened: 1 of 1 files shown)


jpradocueva:
These minor editorial changes will produce a large number of validation errors.
The description of the RR defined in OMNA will differ from the description of Objects already defined.

**Possible solutions:**
* Change the Core LwM2M TS where specifically establishes that the description of a RR cannot be modified.
* or submit a PR including all changes on the Objects affected by this change.

Otherwise, we will have a large number of validation errors, because RR description in OMNA doesn't match with the one inside of the registered Objects.

>See presentation on this topic:  
http://member.openmobilealliance.org/ftp/Public_documents/IPSO/2019/OMA-IPSO-2019-0101-INP_Impact_of_Changes_in_RR_Description.zip

akeranen:
As discussed in the call, probably the best way forward is to add this check (full stops in the end of descriptions) to the validation now and fix descriptions when we do updates to objects anyway.

jpradocueva:
It may be possible to address these editorial issues right at the end of the resolution with the  "units". In this way we can provide an update on "units" and editorial fixes all in one. This may imply to update the Objects files as well, but we don't need to update the Object version. The staff volunteer to do this work as long as the group agrees to do all the work at once.

Issue #185, opened 2019-06-21
[Action Item] Measurement Quality update


seanmcilroy29:
Add a new optional resource ”Measurement Quality” (6042) (R, integer) to the required objects

- [x] Rename the description to Measurement Quality Enumeration and create a new RR called Measurement Quality Percentage that is a range from 0..100

 - [x] Working Group to create a table with all the ID'd Objects and RRs that require to be modified & forward to OMA Staff

  - [x] OMA Staff Implement the changes required to Harmonise resource ”Measurement Quality”

dirtyha:
Updated PR #159 so that the name of RR 6042 is Measurement Quality Indicator as suggested by @mkgillmore

dirtyha:
Should I include the new RR Measurement Quality Percentage into the same PR #159 or create a new PR? What should be the id for the new RR?

seanmcilroy29:
@dirtyha - The next RR id is 6049.
If you're going to register a new RR could you submit a new PR via https://devtoolkit.openmobilealliance.org/OEditor/LWMOEdit

akeranen:
This one is resolved, right?

jpradocueva:
yes this one can be closed. RR 6049 Measurement Quality Level has been registered. 

http://www.openmobilealliance.org/wp/OMNA/LwM2M/LwM2MRegistry.html

seanmcilroy29:
See PR https://github.com/OpenMobileAlliance/lwm2m-registry/pull/516

PR #218, opened 2019-07-31
[Vaisala] Re-usable resource Measurement Quality Level (6049)

dirtyha:
This new RR was requested in action item #185

### DIFF
--- a/Common.xml
+++ b/Common.xml
@@ -3519,13 +3519,13 @@ Using values in range 0-127 is recommended to avoid ambiguities with byte order 
       </Item>
       
       <Item ID="6049">    
-        <Name>Reserved</Name>     
-        <Operations></Operations>     
-        <Type></Type>
-        <RangeEnumeration></RangeEnumeration>
-        <Units></Units>
-        <Submitter></Submitter>
-        <Description></Description>
+        <Name>Measurement Quality Level</Name>     
+        <Operations>R</Operations>     
+        <Type>Integer</Type>
+        <RangeEnumeration>0..100</RangeEnumeration>
+        <Units>%</Units>
+        <Submitter>Vaisala</Submitter>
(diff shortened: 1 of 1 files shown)


Common.xml
```diff
@@ -3519,13 +3519,13 @@ Using values in range 0-127 is recommended to avoid ambiguities with byte order
...
-        <Type></Type>
-        <RangeEnumeration></RangeEnumeration>
-        <Units></Units>
-        <Submitter></Submitter>
-        <Description></Description>
+        <Name>Measurement Quality Level</Name>     
+        <Operations>R</Operations>     
+        <Type>Integer</Type>
+        <RangeEnumeration>0..100</RangeEnumeration>
+        <Units>%</Units>
```
akeranen: I think this is actually better without unit since it’s not strictly percentage. And we don’t have yet unit ID for 0-100% (it’s coming though)

Common.xml
```diff
@@ -3519,13 +3519,13 @@ Using values in range 0-127 is recommended to avoid ambiguities with byte order
...
+        <Submitter>Vaisala</Submitter>
+        <Description><![CDATA[Measurement quality level reported by a smart sensor. Value 100 means that the measured value is good, while value 0 means a rejected value that should not be used. Values between 0 and 100 indicate that the measured value may have quality issues. Smaller values indicate bad quality.]]></Description>
```
akeranen: I wonder if we should further qualify “good” and “bad” here. Is there some more formal language used in the measurement community for these things?

akeranen (changes requested):

dirtyha:
Thanks for the comments Ari. I removed the unit and tried to improve the description. Unfortunately I'm not aware of any formal definition of levels and related languages.

Common.xml
mkgillmore: We recommend removing the unit.  9/19/19

dirtyha: The unit was already removed on Aug 2 commit.

PR #219, opened 2019-07-31
[ICA] Manhole Cover - 3408

QingAn:
Manhole Cover object from ICA

### DIFF
--- a/3408.xml
+++ b/3408.xml
@@ -35,9 +35,9 @@ www.ica-alliance.org
 		<Name>Manhole Cover</Name>
 		<Description1><![CDATA[Describe the status of a manhole cover.]]></Description1>
 		<ObjectID>3408</ObjectID>
-		<ObjectURN></ObjectURN>
+		<ObjectURN>urn:oma:lwm2m:ext:3408</ObjectURN>
 		<LWM2MVersion>1.1</LWM2MVersion>
-		<ObjectVersion />
+		<ObjectVersion>1.0</ObjectVersion>
 		<MultipleInstances>Multiple</MultipleInstances>
 		<Mandatory>Optional</Mandatory>
 		<Resources>
@@ -49,7 +49,7 @@ www.ica-alliance.org
 				<Type>Integer</Type>
 				<RangeEnumeration>0..180</RangeEnumeration>
 				<Units>rad</Units>
-				<Description><![CDATA[The threshold of manhole cover's safety angle]]></Description>
+				<Description><![CDATA[The threshold of manhole cover's safety angle.]]></Description>
 			</Item>
 			<Item ID="2">
 				<Name>Manhole Cover Tilt</Name>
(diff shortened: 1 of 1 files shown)


seanmcilroy29:
@QingAn - Could you update your Object submission with the following;

- Object version 1.0
- Object URN with - urn:oma:lwm2m:ext:3408

3408.xml
```diff
@@ -0,0 +1,79 @@
...
+				<Units>rad</Units>
+				<Description><![CDATA[The tilt angle of a manhole cover]]></Description>
+			</Item>
+			<Item ID="3">
+				<Name>Cover Open Status</Name>
+				<Operations>R</Operations>
+				<MultipleInstances>Single</MultipleInstances>
+				<Mandatory>Mandatory</Mandatory>
+				<Type>Boolean</Type>
+				<RangeEnumeration>Not detected open (0) or Detected open (1)</RangeEnumeration>
```
akeranen: This field can be empty since only two values are accepted: true and false

ScottTaftPotter (commented):
For Safety Angle and Manhole Cover Tilt, the units are radians, but the range enumeration is 0-180 which appears to be degrees.

3408.xml
```diff
@@ -0,0 +1,79 @@
...
+				<Units></Units>
+				<Description><![CDATA[Indicate whether an open status of the manhole cover is detected]]></Description>
```
akeranen: This sounds a bit convoluted. Maybe something like: "Indicates whether the manhole cover is detected open (true) or closed (false)".

3408.xml
```diff
@@ -0,0 +1,79 @@
...
+		<Resources>
+			<Item ID="1">
+				<Name>Safety Angle Threshold</Name>
+				<Operations>R</Operations>
+				<MultipleInstances>Single</MultipleInstances>
+				<Mandatory>Mandatory</Mandatory>
+				<Type>Integer</Type>
+				<RangeEnumeration>0―180</RangeEnumeration>
+				<Units>rad</Units>
+				<Description><![CDATA[The threshold of manhole cover's safety angle]]></Description>
```
akeranen: Please add full stops at the end of descriptions.

3408.xml
```diff
@@ -0,0 +1,79 @@
...
+		<MultipleInstances>Multiple</MultipleInstances>
+		<Mandatory>Optional</Mandatory>
+		<Resources>
+			<Item ID="1">
+				<Name>Safety Angle Threshold</Name>
+				<Operations>R</Operations>
+				<MultipleInstances>Single</MultipleInstances>
+				<Mandatory>Mandatory</Mandatory>
+				<Type>Integer</Type>
+				<RangeEnumeration>0―180</RangeEnumeration>
```
akeranen: If the unit is radians, the range is probably different than 0-180? And type should be float? Or if degrees is really desirable, it's possible to register that as alternative unit in the new SenML units registry.

PR #223, opened 2019-08-20
[B] Easy changes to Units

jpradocueva:
new pull request template

### DIFF
--- a/3396.xml
+++ b/3396.xml
@@ -63,7 +63,7 @@ ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEAL
 				<Mandatory>Mandatory</Mandatory>
 				<Type>String</Type>
 				<RangeEnumeration>1-64 bytes</RangeEnumeration>
-				<Units>byte</Units>
+				<Units>B</Units>
 				<Description><![CDATA[Describes the location of the device within the building. The content of the string is site specific.]]></Description>
 			</Item>
 			<Item ID="5850">
--- a/3396.xml
+++ b/3396.xml
@@ -62,7 +62,7 @@ ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEAL
 				<MultipleInstances>Single</MultipleInstances>
 				<Mandatory>Mandatory</Mandatory>
 				<Type>String</Type>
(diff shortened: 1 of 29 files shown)


jmudge (approved):

seanmcilroy29 (approved):

seanmcilroy29:
Submitted to [R&A](https://member.openmobilealliance.org/Portal_RA/RADetails.asp?RAid=7086&VoteTbId=280&Decision=All&Alone=0&TBId=280&SubTB=280&nbToDisplay=100&Page=1&Param=mUHc53D4BlqrxuxqQinhi) From 2019-09-06 at 08:00 GMT to 2019-09-11 at 18:00 GMT

3396.xml
```diff
@@ -63,7 +63,7 @@ ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEAL
 				<Mandatory>Mandatory</Mandatory>
 				<Type>String</Type>
 				<RangeEnumeration>1-64 bytes</RangeEnumeration>
-				<Units>byte</Units>
+				<Units>B</Units>
```
akeranen, line 66: I think this resource should actually be unitless since the value is an arbitrary string (“description of location in building“).

The enumeration should be “1..64”.

akeranen (approved):
All changes LGTM but the 3396 resource probably has error in the unit.

3396.xml
hannestschofenig: I agree with Ari. The enumeration should say "1..64" and the <Units> element should be empty.

jpradocueva: @hannestschofenig , @akeranen 
 The change to add "1..64" is done but the group still needs to decide if we need to remove the Units.
This Object was submitted by OpenAIS, do we need to check with them this suggestion about unit-less?

seanmcilroy29: PR Agreed

akeranen: The [RangeEnumeration field](http://www.openmobilealliance.org/release/LightweightM2M/V1_1_1-20190617-A/HTML-Version/OMA-TS-LightweightM2M_Core-V1_1_1-20190617-A.html#12-1-0-D1-Object-Template) should not contain word ”Bytes”, just the numbers and delimiters.

jpradocueva: @akeranen we need to evaluate all the Objects from OpenAIS in case other resources have the same problem.

PR #243, opened 2019-10-12
[ADMIN] merge South East Water into test

jpradocueva:
This merge will show errors in RR 6000 and 6001 inside of Object ID 10313. These errors is because in these object the units and range an enumeration was updated to meet the new format of units and range and enumeration expression. We can ignore these errors in this case.
The object passed xml and TS validation (with the above remarks).

### DIFF
--- a/10313.xml
+++ b/10313.xml
@@ -1,0 +1,240 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<!--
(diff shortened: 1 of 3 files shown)


jpradocueva:
@jmudge @seanmcilroy29 we need to arrange a meeting to resolve this conflict. One of the conflict seems related to the CRLF problem.

PR #250, opened 2019-10-15
[ADMIN] Populate DDF.xml ObjectID 10314 entry

jmudge:


jpradocueva:
This PR has failed TS/DDF validation due a mismatch between schema version and LwM2MVersion.

seanmcilroy29 (approved):

jpradocueva (approved):
It seems that the Object itself has some problems, at least with the URN.
The validation should show up that problem.

PR #265, opened 2019-10-29
[Vaisala] Added new RR-6050 Fractional Timestamp.

dirtyha:
Updated also RR-5518 Timestamp description so that it can be used for any measurement, not only location.

### DIFF
--- a/Common.xml
+++ b/Common.xml
@@ -3537,7 +3537,7 @@ Using values in range 0-127 is recommended to avoid ambiguities with byte order 
         <RangeEnumeration>0..1</RangeEnumeration>
         <Units>s</Units>
         <Submitter>Vaisala</Submitter>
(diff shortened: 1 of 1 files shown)


jpradocueva:
**For Information**
Changes to RR 5518 will affect the following Objects: 10311 & 3336.

dirtyha:
> **For Information**
> Changes to RR 5518 will affect the following Objects: 10311 & 3336.

True, maybe we should create a separate PR to update those objects that repeat RR 5518 (Timestamp) description. Generally speaking, it looks like there are plenty of other description mismatches between objects using RR and RR definition. Maybe a script could be created to find these out.

jpradocueva:
@dirtyha the LwM2M Editor Tool has a validation tool and provides an update on any description mismatches. The problem is that we are trying to resolve the issues related to the Units and this is creating a backlog of pending issues. Once it is resolved we can even make the validation tool publicly available.

PR #267, opened 2019-10-31
Range fixes to 102?? objects; includes whitespace fixes

akeranen:
Fixing range delimiter for all remaining objects.


akeranen:
This fix was done automatically without checking the descriptions or other semantics of the resources. Note that the ".." range is [defined to include both the lower and upper bound](http://www.openmobilealliance.org/release/LightweightM2M/V1_1_1-20190617-A/HTML-Version/OMA-TS-LightweightM2M_Core-V1_1_1-20190617-A.html#12-1-0-D1-Object-Template).

PR #268, opened 2019-10-31
Using /100 instead of %

akeranen:
Aligning the percentage unit use with the new SenML registry entry


jpradocueva:
This unit is still pending to be added to SenML registry. Therefore, there is no point to validate the PR as it will indicate errors.

akeranen:
"/100" is now in the latest [more-units-draft](https://tools.ietf.org/html/draft-ietf-core-senml-more-units-03#page-5).

PR #319, opened 2019-11-06
[CloudMinds] 10365- Gas-Dust Sensor

Kenny-SZ:
This LWM2M Object includes Gas-Dust information.

### DIFF
--- a/10365.xml
+++ b/10365.xml
@@ -1,0 +1,57 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!--BSD 3-Clause License
+
+Copyright (c) 2019, Cloudminds
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are met:
+
+* Redistributions of source code must retain the above copyright notice, this
+  list of conditions and the following disclaimer.
+
+* Redistributions in binary form must reproduce the above copyright notice,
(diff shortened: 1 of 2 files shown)


10365.xml
```diff
@@ -0,0 +1,57 @@
...
+    <Name>Gas-Dust Sensor</Name>
+    <Description1><![CDATA[This LWM2M Object includes Gas-Dust information.]]></Description1>
+    <ObjectID>10365</ObjectID>
+    <ObjectURN>urn:oma:lwm2m:x:10365</ObjectURN>
+    <LWM2MVersion>1.0</LWM2MVersion>
+    <ObjectVersion>1.0</ObjectVersion>
+    <MultipleInstances>Single</MultipleInstances>
+    <Mandatory>Optional</Mandatory>
+    <Resources>
+		<Item ID="50">
```
akeranen, line 44: Instead of having separate objects with the same structure (just status resource), it would make more sense to have single object that has the status resource and another resource that describes for **what** the status is given.

Kenny-SZ: Hi Akeranen, thanks for your comments, sorry for late reply, since it has been flooded by lots of comments of objects(common object or other company's).

Here we only list very basic resources, but later on, it may include more specific resources or operation(E). Thanks.

mkgillmore:

PR #320, opened 2019-11-06
[CloudMinds] 10366-Fan

Kenny-SZ:
This LWM2M Object includes Fan information.

### DIFF
--- a/10366.xml
+++ b/10366.xml
@@ -1,0 +1,59 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!--BSD 3-Clause License
+
+Copyright (c) 2019, Cloudminds
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are met:
+
+* Redistributions of source code must retain the above copyright notice, this
+  list of conditions and the following disclaimer.
+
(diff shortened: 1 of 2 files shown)


10366.xml
```diff
@@ -0,0 +1,59 @@
...
+    <Name>Fan</Name>
+    <Description1><![CDATA[This LWM2M Object includes Fan information.]]></Description1>
+    <ObjectID>10366</ObjectID>
+    <ObjectURN>urn:oma:lwm2m:x:10366</ObjectURN>
+    <LWM2MVersion>1.0</LWM2MVersion>
+    <ObjectVersion>1.0</ObjectVersion>
+    <MultipleInstances>Multiple</MultipleInstances>
+    <Mandatory>Optional</Mandatory>
+    <Resources>
+		<Item ID="1">
```
akeranen, line 44: For a generic Fan object seems that much more (optional) information would make sense. For example fan speed?

akeranen (changes requested):

10366.xml
Kenny-SZ: Hi Akeranen,
     Thanks for your comments, Fan speed is defined in 10358(Fan PM), and I believe we will add more resources for Fan and other Objects.

PR #321, opened 2019-11-06
[CloudMinds] 10335-Event

Kenny-SZ:
Here lists all items of an event reported by the robot.

### DIFF
--- a/10335.xml
+++ b/10335.xml
@@ -1,0 +1,97 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!--BSD 3-Clause License
+
+Copyright (c) 2019, Cloudminds
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are met:
+
+* Redistributions of source code must retain the above copyright notice, this
+  list of conditions and the following disclaimer.
+
+* Redistributions in binary form must reproduce the above copyright notice,
+  this list of conditions and the following disclaimer in the documentation
+  and/or other materials provided with the distribution.
+
+* Neither the name of the copyright holder nor the names of its
+  contributors may be used to endorse or promote products derived from
+  this software without specific prior written permission.
+
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+-->
+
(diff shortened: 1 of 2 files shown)


10335.xml
```diff
@@ -0,0 +1,97 @@
...
+DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+-->
+
+<LWM2M xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="http://openmobilealliance.org/tech/profiles/LWM2M.xsd">
```
seanmcilroy29, line 33: Invalid XML Schema ref as you've indicated ```<LWM2MVersion>1.1</LWM2MVersion>``` 
Schema should be ```<LWM2M xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="http://www.openmobilealliance.org/tech/profiles/LWM2M-v1_1.xsd">```

jpradocueva: @Kenny-SZ do you really need Version 1.1? Based on the "Type" used in your object you can use the schema defined for Version 1.0 of the LwM2M protocol.

Kenny-SZ: Thanks for your comments, 1.0 is enough.

10335.xml
```diff
@@ -0,0 +1,97 @@
...
+    <Resources>
+		<Item ID="1">
+			<Name>Entity</Name>
+			<Operations>R</Operations>
+			<MultipleInstances>Single</MultipleInstances>
+			<Mandatory>Mandatory</Mandatory>
+			<Type>String</Type>
+			<RangeEnumeration>4..63</RangeEnumeration>
+			<Units/>
+			<Description><![CDATA[Contains the object ID and object instance ID/name, for example: &lt;/10320/2&gt;.]]></Description>
```
akeranen, line 52: This sounds like Object Link type?

10335.xml
```diff
@@ -0,0 +1,97 @@
...
+		</Item>
+		<Item ID="2">
+			<Name>Event Type</Name>
+			<Operations>R</Operations>
+			<MultipleInstances>Single</MultipleInstances>
+			<Mandatory>Mandatory</Mandatory>
+			<Type>Integer</Type>
+			<RangeEnumeration>0..65535</RangeEnumeration>
+			<Units></Units>
+			<Description><![CDATA[The type of this event, for example: APP restart.]]></Description>
```
akeranen, line 62: How are the numbers mapped to types?

10335.xml
```diff
@@ -0,0 +1,97 @@
...
+		</Item>
+		<Item ID="3">
+			<Name>Time</Name>
+			<Operations>R</Operations>
+			<MultipleInstances>Single</MultipleInstances>
+			<Mandatory>Mandatory</Mandatory>
+			<Type>Time</Type>
+			<RangeEnumeration></RangeEnumeration>
+			<Units></Units>
+			<Description><![CDATA[When was this event reported.]]></Description>
```
akeranen, line 72: There's also re-usable time stamp resource that could be used here

akeranen (changes requested):

10335.xml
Kenny-SZ: Yes, if it's object ID + instance ID，then it likes a Objlnk, but when it's Object ID + instance name， then it's not a Objlnk.

For some objects, it has a lot of instances, and the instance ID is a little meaningless, we tends to use instance name to indicate a object instance.
For some objects, we don't define a OMA object for them, we may use  /Computer/Kenny to indicate Kenny's Computer in an event.

10335.xml
Kenny-SZ: We have defined dozens of event types, but I didn't list them here.

For example: 3100 is for APP restart, 5000 is for robot restart.

In my opinion, it's not a good way to use one Object ID for each Event or Alarm, we prefer to use only one Object for Event, and use its resource to indicate which Event/Alarm it is.

10335.xml
Kenny-SZ: If there is already a reused resource, and the meaning is the same, we can consider to use it, thanks.

mkgillmore:

PR #322, opened 2019-11-06
[CloudMinds] 10362-Drop Sensor

Kenny-SZ:
This LWM2M Object includes Drop Sensor information.

### DIFF
--- a/10362.xml
+++ b/10362.xml
@@ -1,0 +1,91 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!--BSD 3-Clause License
+
+Copyright (c) 2019, Cloudminds
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are met:
+
+* Redistributions of source code must retain the above copyright notice, this
+  list of conditions and the following disclaimer.
+
+* Redistributions in binary form must reproduce the above copyright notice,
+  this list of conditions and the following disclaimer in the documentation
+  and/or other materials provided with the distribution.
+
+* Neither the name of the copyright holder nor the names of its
+  contributors may be used to endorse or promote products derived from
+  this software without specific prior written permission.
+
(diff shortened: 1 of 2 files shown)


10362.xml
```diff
@@ -0,0 +1,91 @@
...
+        </Item>
+		<Item ID="51">
+			<Name>Drop Detection</Name>
+			<Operations>R</Operations>
+			<MultipleInstances>Single</MultipleInstances>
+			<Mandatory>Optional</Mandatory>
+			<Type>Boolean</Type>
+			<RangeEnumeration/>
+			<Units/>
+			<Description><![CDATA[If Drop Detection is enabled or not,
```
akeranen, line 74: Boolean values are "true" and "false"

10362.xml
```diff
@@ -0,0 +1,91 @@
...
+		
+	  	<Item ID="100">
+			<Name>Drop Detection Control</Name>
+			<Operations>E</Operations>
+			<MultipleInstances>Single</MultipleInstances>
+			<Mandatory>Mandatory</Mandatory>
+			<Type/>
+			<RangeEnumeration/>
+			<Units/>
+			<Description><![CDATA[Enable/Disable Drop Detection: Parameter: 1:Enable, 0:Disable]]></Description>
```
akeranen, line 86: Should this be type boolean with true/false?

akeranen (changes requested):

10362.xml
Kenny-SZ: Yes, changed. Thanks.

10362.xml
Kenny-SZ: Currently true for enable, false for not.

10362.xml
seanmcilroy29: Incorrect Type/Operation combination - A Resource, which supports the "Execute" operation, MUST have no associated Data Type (none) encoded as an empty value in Object DDF file

Kenny-SZ: Removed, thanks.

mkgillmore:

PR #340, opened 2019-11-19
[ADMIN] IPSO license update

seanmcilroy29:
IPSO license update

### DIFF
--- a/3200.xml
+++ b/3200.xml
@@ -2,7 +2,7 @@
 
 <!-- BSD-3 Clause License
 
-Copyright 2019 Open Mobile Alliance. All rights reserved.
+Copyright 2019 Open Mobile Alliance. 
 
 Redistribution and use in source and binary forms, with or without
 modification, are permitted provided that the following conditions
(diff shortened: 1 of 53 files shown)


jpradocueva (changes requested):
The name of the organization should be kept to IPSO.

seanmcilroy29:
If IPSO has been dissolved as an Alliance to become a WG entity within OMA, shouldn't OMA be the over arching Copy right organisation?

jpradocueva:
This is a tough call. My view is that the objects were created by IPSO. If we now change the name of the organization we are overwriting the original creator. I don't think that this is right. These objects were created outside of OMA.

jmudge:
Does IPSO still exist as a legal entity now? My impression is that copyright can only be asserted by legal entities. (Note the copyright year is being updated to 2019.)

seanmcilroy29:
I've created a PR to update these Objects with IPSO. 
When these Objects are submitted with the updated BSD license to ODM, IPSO will be highlighted as holding the Copyright, but is this correct?

PR #362, opened 2019-11-29
[ADMIN] new file that contains additional senml units.

jpradocueva:
This file will contain senml units in addition to the ones provided in the senml registry: https://www.iana.org/assignments/senml/senml.xhtml

new pull request template


jpradocueva:
this document is a placeholder to store additional senml units. At this moment the file is empty as the content will be added later.

PR #374, opened 2019-12-05
[ADMIN] merge "Odins" branch into "Revised-Common"

jpradocueva:
All the changes related to the Units and CRLF should be merged first into Revised-Common and from there will be merged into "test" when ready.


seanmcilroy29 (approved):
This PR also merges PRs that have previously been merged into the Test branch

jpradocueva:
@seanmcilroy29 it should be ok if the same content has been merged already into the "test" branch. By having everything in the "Revised-Commo" we can what is still pending and we can decide if we can merge "test" into "prod".

PR #375, opened 2019-12-06
[ADMIN] file with additional senml unitls

jpradocueva:
This file contains additional senml units that are not currently registered with SenML Registry.
The file may be used during the validation, an optional tick box will allow to select or not select the file as well as to indicate where the file is located.

Also a new schema called "LWM2M_senml_units.xsd" has also being created to ensure that any changes to this file are correct.
This file "LWM2M_senml_units" is added to the repository to place under version control.

### DIFF
--- a/LWM2M_senml_units.xml
+++ b/LWM2M_senml_units.xml
@@ -1,0 +1,157 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<units     xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="http://openmobilealliance.org/tech/profiles/LWM2M_senml_units.xsd">
+  <title>Additional SenML Units</title>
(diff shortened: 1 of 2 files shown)


seanmcilroy29 (commented):
Should the following units be added;

   | B      | Byte (information content)           
   | VA     | volt-ampere (Apparent Power)         
   | var    | volt-ampere reactive (Reactive Power)                               
   | J/m    | joule per meter (Energy per distance)

jpradocueva:
Those units are already in the SenML Registry, that this the reason why I haven't added to the additional LWM2M-senml-units  file. See below:

image

seanmcilroy29:
I was using https://tools.ietf.org/html/rfc8428 and https://tools.ietf.org/html/draft-ietf-core-senml-more-units-02#page-5 SenML Units Registry as a reference, which on review differs from https://www.iana.org/assignments/senml/senml.xhtml as this has a mix of both

jmudge (approved):

PR #380, opened 2019-12-11
[ADMIN] Add ID to the registry element

jpradocueva:
The file LWM2M_senml_units allows to store additional senml-units that are used for the validation of Objects submitted to the LwM2M Registry.


seanmcilroy29 (approved):
In GitKraken, It looks like each line ending was updated but it's not clear with what

jmudge (commented):
Note git is indicating that the file LWM2M_senml_units.xml is changed but not staged.

PR #386, opened 2019-12-12
[Bug] OpenAIS unit & Range Enumeration updated

seanmcilroy29:
Unit Ws updated to J in Objects 3388, 3396, 3397, 3399
Associated descriptions updated with ``J is equivalent to Watt-seconds.``
- Unit update and description agreed by Walter Werner <werner@werner-ms.at> 

OpenAIS Objects Range Enumeration updated from ``-`` to ``..``

**Validation removes Ws error**


jmudge (approved):

jpradocueva (approved):
ok.

PR #387, opened 2019-12-12
[Bug] OpenAIS Range Enumeration updated

seanmcilroy29:
Unit Ws updated to J in Objects 3388, 3396, 3397, 3399
Associated descriptions updated with **J is equivalent to Watt-seconds.**
- Unit update and description agreed by Walter Werner werner@werner-ms.at

OpenAIS Objects Range Enumeration updated from ``- ``to ``..``


mkgillmore:
Group agrees to merge 12/19/19 with the exception to change unit of J back to Ws.  Request to register Ws in SenML units.

seanmcilroy29:
Update - OpenAIS Objects Range Enumeration updated from - to .. only

PR #395, opened 2019-12-18
[ADMIN] oneM2M Objects updated with BSD3 License

seanmcilroy29:
BSD3 License added to oneM2M Objects

Agreed by Roland Hechwartner roland.hechwartner@magenta.at - oneM2M TP Chair

Validated with no errors


jpradocueva (approved):
Ok, all validates correctly.

Issue #184, opened 2019-06-21
[Action Item] Add new optional resource 'Timestamp'


seanmcilroy29:
Create PR to add new optional resources ”Timestamp” (5518) to all sensor objects
- Application can then explicitly read measurement time provided in the object model

dirtyha:
I propose that optional resource 'Timestamp' is added to IPSO smart objects 3200-3350 which represent a sensor:

3200-3203 (analog and digital input/output where a sensor can be connected)
3300-3305
3313-3328
3330-3332
3334-3336
3346

The excluded are objects that represent a controller, actuator or something similar to change device state. I'm not sure if "Timestamp" is needed for them, at least I see no need for us (Vaisala). Maybe for some applications the timestamp could tell when the state was actually changed in the device initiated by the server or done at the device from UI.

akeranen:
Seems to be resolved by #403

seanmcilroy29:
See - https://github.com/OpenMobileAlliance/lwm2m-registry/pull/516

Issue #181, opened 2019-06-20
[Action Item] Harmonise resource 'Application Type' to all sensor objects.


seanmcilroy29:
Harmonise resource 'Application Type' (5750) for all sensor objects.
- [ ] Working Group to create a table with all the ID'd Objects and RRs that require to be modified & forward to OMA Staff
- [ ] OMA Staff Implement the changes required to Harmonise resource 'Application Type'

**Note: Required to be implemented by end Sept 2019**

seanmcilroy29:
Info moved from Issue #160 
A call with @akeranen revealed that adding re-usable resource (Application Type) implicitly to IPSO object is not possible. So, my proposal is to explicitly add it to IPSO objects by creating a new object version.

The following IPSO objects are missing Application Type (5750) reusable resource, possibly accidentally.

ID	Name
3301	Illuminance
3302	Presence
3303	Temperature
3304	Humidity
3305	Power Measurement
3310	Load Control
3313	Accelerometer
3314	Magnetometer
3315	Barometer

akeranen:
Also addressed by #403

seanmcilroy29:
See PR - https://github.com/OpenMobileAlliance/lwm2m-registry/pull/516

PR #402, opened 2020-01-08
[Vaisala] New resources into IPSO sensor objects

dirtyha:
Added new resources specified in issues #181,#184 and #185. These resources are:

1. Application Type (5750) if missing
2. Timestamp (5518) and Fractional Timestamp (6050) if applicable
3. Measurement Quality Indicator (6042) and Measurement Quality Level (6049) if applicable

### DIFF
--- a/10311.xml
+++ b/10311.xml
@@ -40,7 +40,7 @@ POSSIBILITY OF SUCH DAMAGE.
         <ObjectID>10311</ObjectID>
         <ObjectURN>urn:oma:lwm2m:x:10311</ObjectURN>
 		<LWM2MVersion>1.0</LWM2MVersion>
-		<ObjectVersion>1.0</ObjectVersion>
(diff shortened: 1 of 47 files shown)


jpradocueva:
@dirtyha unfortunately, in the current repository if a new version of an Object is needed then, a new file has to be submitted, e.g. if file ObjectID 3200.xml is updated with new optional resources, then the following changes are required:
* update the file name to 3200-1_1.xml
  * in this repository is not possible to store different version of the same file and keep exactly the same file name. This  is due to how the repository was created in the first place. Please refer to the Objects produced by OMA, you will see that each new version of an Object has a new file, one for each version.
* update the URN to "urn:oma:lwm2m:ext:3200:1.1"
* update the "LWM2MVersion" element to 1.1
* add the new optional Resources. If the resources are mandatory, then the changes are considered non-backwards compatible and the version of the Object must be v2.0.

PR #403, opened 2020-01-09
[Vaisala] New optional resources added into IPSO sensor objects

dirtyha:
Added new resources specified in issues #181,#184 and #185. These resources are:

1. Application Type (5750) if missing
2. Timestamp (5518) and Fractional Timestamp (6050) if applicable
3. Measurement Quality Indicator (6042) and Measurement Quality Level (6049) if applicable

Changes have been generated by using bulk_update_tool and [AddResources.py](https://github.com/OpenMobileAlliance/bulk-update-tool/tree/master/vaisala-issue-%23181-%23184-%23185) script.

### DIFF
--- a/10311-1_1.xml
+++ b/10311-1_1.xml
@@ -1,0 +1,194 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<!-- BSD-3 Clause License
+
+Copyright 2019 Vaisala. 
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions
+are met:
+
+1. Redistributions of source code must retain the above copyright
+notice, this list of conditions and the following disclaimer.
+2. Redistributions in binary form must reproduce the above copyright
+notice, this list of conditions and the following disclaimer in the
+documentation and/or other materials provided with the distribution.
+3. Neither the name of the copyright holder nor the names of its
+contributors may be used to endorse or promote products derived
+from this software without specific prior written permission.
+
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+"AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+POSSIBILITY OF SUCH DAMAGE.
+-->
+
+<LWM2M  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="http://openmobilealliance.org/tech/profiles/LWM2M.xsd">
+	<Object ObjectType="MODefinition">
+		<Name>Solar Radiation</Name>
+		<Description1>
+            This object is used to report solar irradiance (SI), i.e. power per unit area received from the Sun in the form of electromagnetic radiation, on a planar surface measured by a pyranometer or similar instrument. A pyranometer measures solar irradiance from the hemisphere above within a wavelength range 0.3 μm to 3 μm. For example, the application of solar radiation measurement can be meteorological networks and solar energy applications.
+        </Description1>
+		<ObjectID>10311</ObjectID>
+		<ObjectURN>urn:oma:lwm2m:x:10311:1.1</ObjectURN>
+		<LWM2MVersion>1.0</LWM2MVersion>
+		<ObjectVersion>1.1</ObjectVersion>
+		<MultipleInstances>Multiple</MultipleInstances>
+		<Mandatory>Optional</Mandatory>
+		<Resources>
+			<Item ID="5601">
+				<Name>Min Measured Value</Name>
+				<Operations>R</Operations>
+				<MultipleInstances>Single</MultipleInstances>
+				<Mandatory>Optional</Mandatory>
+				<Type>Float</Type>
+				<RangeEnumeration></RangeEnumeration>
+				<Units></Units>
+				<Description>
+                    The minimum value measured by the sensor since it is ON or Reset, expressed in the unit defined by the "Sensor Units" resource if present.
+                </Description>
+			</Item>
+			<Item ID="5602">
+				<Name>Max Measured Value</Name>
+				<Operations>R</Operations>
+				<MultipleInstances>Single</MultipleInstances>
+				<Mandatory>Optional</Mandatory>
+				<Type>Float</Type>
+				<RangeEnumeration></RangeEnumeration>
+				<Units></Units>
+				<Description>
+                    The maximum value measured by the sensor since it is ON or Reset, expressed in the unit defined by the "Sensor Units" resource if present.
+                </Description>
+			</Item>
+			<Item ID="5603">
+				<Name>Min Range Value</Name>
+				<Operations>R</Operations>
+				<MultipleInstances>Single</MultipleInstances>
+				<Mandatory>Optional</Mandatory>
+				<Type>Float</Type>
+				<RangeEnumeration></RangeEnumeration>
+				<Units></Units>
+				<Description>
+                    The minimum value that can be measured by the sensor, expressed in the unit defined by the "Sensor Units" resource if present.
+                </Description>
+			</Item>
+			<Item ID="5604">
+				<Name>Max Range Value</Name>
+				<Operations>R</Operations>
+				<MultipleInstances>Single</MultipleInstances>
+				<Mandatory>Optional</Mandatory>
+				<Type>Float</Type>
+				<RangeEnumeration></RangeEnumeration>
+				<Units></Units>
+				<Description>
+                    The maximum value that can be measured by the sensor, expressed in the unit defined by the "Sensor Units" resource if present.
+                </Description>
+			</Item>
+			<Item ID="5605">
+				<Name>Reset Min and Max Measured Values</Name>
+				<Operations>E</Operations>
+				<MultipleInstances>Single</MultipleInstances>
+				<Mandatory>Optional</Mandatory>
+				<Type></Type>
+				<RangeEnumeration></RangeEnumeration>
+				<Units></Units>
+				<Description>
+                    Reset the Min and Max Measured Values to current value.
+                </Description>
+			</Item>
+			<Item ID="5518">
+				<Name>Timestamp</Name>
+				<Operations>R</Operations>
+				<MultipleInstances>Single</MultipleInstances>
+				<Mandatory>Optional</Mandatory>
+				<Type>Time</Type>
+				<RangeEnumeration></RangeEnumeration>
+				<Units></Units>
+				<Description>The timestamp of when the measurement was performed.</Description>
+			</Item>
+			<Item ID="5700">
+				<Name>Sensor Value</Name>
+				<Operations>R</Operations>
+				<MultipleInstances>Single</MultipleInstances>
+				<Mandatory>Mandatory</Mandatory>
+				<Type>Float</Type>
+				<RangeEnumeration></RangeEnumeration>
+				<Units></Units>
+				<Description>Last or Current Measured Value from the Sensor expressed in the unit defined by the "Sensor Units" resource if present.</Description>
+			</Item>
+			<Item ID="5701">
+				<Name>Sensor Units</Name>
+				<Operations>R</Operations>
+				<MultipleInstances>Single</MultipleInstances>
+				<Mandatory>Optional</Mandatory>
+				<Type>String</Type>
+				<RangeEnumeration></RangeEnumeration>
+				<Units></Units>
+				<Description>
+                    Measurement Units Definition.
+                </Description>
+			</Item>
+			<Item ID="5750">
+				<Name>Application Type</Name>
+				<Operations>RW</Operations>
+				<MultipleInstances>Single</MultipleInstances>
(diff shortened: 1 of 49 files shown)


jpradocueva:
My comment of increase the Object Versioning in this PR is based in the current statement in the TS Core, section 7.2. Object Versioning
Based on the extract below Evolution of type II implies, that when a new "Optional" resource is added or removed the Object Version type II is increased, in this case it should be v1.1.

It is ok for the group to discuss if this functionality should be kept in the future but at the moment that is that the TS indicates. We have been applying this rule since was implemented a few years ago.



image

Note: unfortunately the OMA LwM2M Registry was created a few years ago, at that time it was decided that each object version will have its own file name and the object should be added to the file name. In this case, and until the group decides what to do with the Object Version, the repository requires a new file each time that the version is changed.

Hope explanation it is clear. These rules were set up by DMSE and IPSO. The staff is just enforcing its implementation, at least until there are changed.

seanmcilroy29:
DDF File updated with Object version increase

- [x] Editor validation - Passed with no errors
- [x] Validated as well-formed

jpradocueva:
@dirtyha @seanmcilroy29 
This PR also contains new versions for  ObjectIDs: 
- 3327 to 3338 and
- 3342, 3346 to 3350

but the DDF has not been updated to introduce this new revisions. The PR cannot be approved as it is as the above revisions are not listed in the DDF.

seanmcilroy29:
- [x] Omitted Objects added
- [x] Editor Validated with no errors
- [x] Validated as well-formed

dnav:
Two remarks regarding the timestamp resources (ID: 5518 and ID:6050)  being added to all IPSO Objects:

- In the data encoding formats SenML JSON, SenML CBOR, and LwM2M JSON, a timestamp can be attached to a value. So I found the timestamp resources redundant.

- (On this one, I guess the train has left the station) The "Fractional Timestamp" RR (ID: 6050) is a Float between 0 and 1. Floating-point computations can be quite expensive on a constrained system. The usual practice is to use an Integer to store the fixed-point decimal part. See for instance the [UNIX timeval structure](http://man7.org/linux/man-pages/man2/gettimeofday.2.html#DESCRIPTION) which uses microseconds or the [UNIX timespec structure](http://man7.org/linux/man-pages/man2/clock_settime.2.html#DESCRIPTION) which uses nanoseconds.

dirtyha:
> * In the data encoding formats SenML JSON, SenML CBOR, and LwM2M JSON, a timestamp can be attached to a value. So I found the timestamp resources redundant.

My understanding of the specification is that SenML timestamp can only be used in solicited messages (notify & send), not in read replies. Although Vaisala's preference is indeed to use timestamps in SenML messages, there needs to be an explicit timestamp resource for reads too. The second reason is the set of currently available LwM2M server implementations which tend to offer a north bound API for LwM2M objects, not SenML messages, which results the absence of the timestamp.

dnav:
> My understanding of the specification is that SenML timestamp can only be used in solicited messages (notify & send), not in read replies. 

Reading the LwM2M TS, I can understand the confusion. But I can't see anything preventing the usage of timestamps in a READ response.

Actually during the latest testfest, several Client implementation were using them in READ responses.

akeranen:
For easier review, is there (or can someone make) a PR against the previous versions of the object? Would be nice to see just the diff.

dirtyha:
> For easier review, is there (or can someone make) a PR against the previous versions of the object? Would be nice to see just the diff.

Ari, there is an old PR #402 that has the same content against previous object versions.

dirtyha:
> Reading the LwM2M TS, I can understand the confusion. But I can't see anything preventing the usage of timestamps in a READ response.
> 
> Actually during the latest testfest, several Client implementation were using them in READ responses.

Thanks, this is interesting. I would still keep Timestamp and Fractional Timestamp in the PR. They are not mandatory for the client to implement if the LwM2M server has the capability to work with timestamps inside SenML messages. Otherwise, if the LwM2M server implementation is unknown or there is a chance that message format is not a SenML variant, then the device manufacturer can provide explicit timestamp of the sensor measurement inside the object model.

10311-1_1.xml
```diff
@@ -0,0 +1,194 @@
...
+			</Item>
+			<Item ID="6042">
+				<Name>Measurement Quality Indicator</Name>
+				<Operations>R</Operations>
+				<MultipleInstances>Single</MultipleInstances>
+				<Mandatory>Optional</Mandatory>
+				<Type>Integer</Type>
+				<RangeEnumeration>0..23</RangeEnumeration>
+				<Units></Units>
+				<Description>Measurement quality indicator reported by a smart sensor. 0: UNCHECKED No quality checks were done because they do not exist or can not be applied. 1: REJECTED WITH CERTAINTY The measured value is invalid. 2: REJECTED WITH PROPABILITY The measured value is likely invalid. 3: ACCEPTED BUT SUSPICIOUS The measured value is likely OK. 4: ACCEPTED The measured value is OK. 5-15: Reserved for future expansion. 16-23: Vendor specific measurement quality.</Description>
```
jaimejim: This is quite a strange thing to carry around as a measurement. I'd assume that the application logic needs to decide what to do with the measurement based on the value of 6049. I don't see why would a lwm2m client ever tell a lwm2m server what to do with the measurement, like "ACCEPTED BUT SUSPICIOUS" that is application logic, that is not a measurement.

akeranen:
Editorial issues:
* There's a typo with word PROPABILITY (PROBABILITY) in the description of the resource 6042
* Instead of "Reserved for future expansion" probably better would be "Reserved for future extensions"

3200-1_1.xml
```diff
@@ -0,0 +1,148 @@
...
+			</Item>
+			<Item ID="6050">
+				<Name>Fractional Timestamp</Name>
+				<Operations>R</Operations>
+				<MultipleInstances>Single</MultipleInstances>
+				<Mandatory>Optional</Mandatory>
+				<Type>Float</Type>
+				<RangeEnumeration>0..1</RangeEnumeration>
+				<Units>s</Units>
+				<Description>Fractional part of the timestamp when sub-second precision is used (e.g., 0.23 for 230 ms).</Description>
```
jaimejim, line 143: This is weird. I am assuming that the case is a lwm2m client sending the current time (e.g., 1582642022), by adding fractional time we add subsecond precission (e.g., 0.23 if it is 230 ms).
I think Fractional Time should be the current time in ms (e.g., 1582642022.23) and that it should be used instead of Timestamp. Or just use integer and say that it is time in ms.

There is also the issue that senml also does send timestamps, so you may end up with redundant information.

jaimejim (changes requested):
I have some comments that I think could be good to address. They are about resource 6050 and 6042.  Sorry for the delay.

3200-1_1.xml
akeranen: Only fractional (floating point) time doesn't work well since not all platforms support float types. Also just integer (fixed point) does not work well since we would have to fix the precision (ms/us/ns?) in the resource definition. 

One potential alternative would be to create a new "Float timestamp" resource that could be used when float is supported, and could include the non-fractional part of the time too. But that should not be used together with the timestamp or the fractional part resources, or we run into weird error cases. Hence, that might not be worth the effort.

For SenML time stamp configuration a new attribute might be a good solution.

10311-1_1.xml
akeranen: IIRC, this metadata about the measurement was for more advanced sensor types that have the sufficient logic to know more about the measurement quality than the LwM2M server would know (e.g., based only on 6049).

3201-1_1.xml
```diff
@@ -0,0 +1,98 @@
...
+			</Item>
+			<Item ID="5518">
+				<Name>Timestamp</Name>
+				<Operations>R</Operations>
+				<MultipleInstances>Single</MultipleInstances>
+				<Mandatory>Optional</Mandatory>
+				<Type>Time</Type>
+				<RangeEnumeration></RangeEnumeration>
+				<Units></Units>
+				<Description>The timestamp of when the measurement was performed.</Description>
```
akeranen, line 83: Does the timestamp resource make sense for actuators like digital output?

3203-1_1.xml
```diff
@@ -0,0 +1,128 @@
...
+			</Item>
+			<Item ID="5518">
+				<Name>Timestamp</Name>
+				<Operations>R</Operations>
+				<MultipleInstances>Single</MultipleInstances>
+				<Mandatory>Optional</Mandatory>
+				<Type>Time</Type>
+				<RangeEnumeration></RangeEnumeration>
+				<Units></Units>
+				<Description>The timestamp of when the measurement was performed.</Description>
```
akeranen, line 93: Also here: timestamp for actuator?

3203-1_1.xml
```diff
@@ -0,0 +1,128 @@
...
+				<Operations>R</Operations>
+				<MultipleInstances>Single</MultipleInstances>
+				<Mandatory>Optional</Mandatory>
+				<Type>Float</Type>
+				<RangeEnumeration>0..1</RangeEnumeration>
+				<Units>s</Units>
+				<Description>Fractional part of the timestamp when sub-second precision is used (e.g., 0.23 for 230 ms).</Description>
+			</Item>
+			<Item ID="6042">
+				<Name>Measurement Quality Indicator</Name>
```
akeranen: Also measurement quality for actuators sounds a bit odd.

akeranen (changes requested):

akeranen:
Probably we should not have measurement quality or time stamp resources for the actuator objects. Or what would be the semantics for these for something the LwM2M MS writes? I commented for a few above, but also likely applies to 3308, 3310, 3312, and 3350 (and maybe others?).

10311-1_1.xml
dirtyha: There are multiple use cases where a weather instrument may flag a measurement with "ACCEPTED BUT SUSPICIOUS". In general, quality control checks are done in multiple levels. The first level checks are done already at the smart instrument. For example, a multi-sensor weather instrument may detect that precipitation type is snow but air temperature is +5C. This is likely a failure in precipitation type detection but can be possible in some rare cases, so it makes sense to tag the precipitation type measurement, perhaps also the temperature measurement. I'm not a meteorologist, so do not take +5C as an important threshold, but you probably get what I'm after.

3200-1_1.xml
dirtyha: > There is also the issue that senml also does send timestamps, so you may end up with redundant information.

This was already commented earlier by @dnav, so check my comment above. Since the LwM2M specification does not specify LwM2M server north bound interface, it depends on the server implementation if senml timestamps are available via nb interface for applications.

3201-1_1.xml
dirtyha: It is the time when the digital output state changed. If the state change was triggered by the device then the server doesn't know it until it reads the sate or receives a send or notification message.

3203-1_1.xml
dirtyha: See the response above.

3203-1_1.xml
dirtyha: Object 3203 is Analog Output. My reason for adding Measurement Quality is that the output voltage requires a measurement that involves a level of uncertainty that can be reported. I admit that this was something that I hesitated to add, but did it anyway.

Btw, the resource 5650 Analog Output Current Value has likely a bug in range (0..1). The counterpart (Analog Inpout) resource 5600 Analog Input Current Value does not have range limitation.

dirtyha:
> Probably we should not have measurement quality or time stamp resources for the actuator objects. Or what would be the semantics for these for something the LwM2M MS writes? I commented for a few above, but also likely applies to 3308, 3310, 3312, and 3350 (and maybe others?).

3308 Set Point: timestamp applicable by using the same logic as with Analog Output, i.e. when the state was changed.
3310 Load Control: timestamp applicable (the same as above)
3312 Power Control: timestamp applicable (the same as above)
3350 Stopwatch: timestamp & quality applicable, example is precipitatoin duration measurement. I.e. it has been raining 3660 seconds when last measured at 2020-02-26T09:43:11Z. Quality applies as well, i.e. OK normally, but can be something else depending on the situation.

3201-1_1.xml
akeranen: In that case the description is a bit misleading (I.e., the resource is not a good fit). If we really want this functionality, maybe it’s better to define a new resource that has better fitting semantics (description) for this?

Or can we make a semantically compatible change to the RR description? “Time when change in the value was detected”? Sounds a bit borderline case though.

akeranen: In that case the description is a bit misleading (I.e., the resource is not a good fit). If we really want this functionality, maybe it’s better to define a new resource that has better fitting semantics (description) for this?

3200-1_1.xml
akeranen: Any reasonable LwM2M server implementation should expose all information available from the clients. I guess the problem here is that using SenML time stamps was never properly defined in the LwM2M TS. That could be fixed in future versions. 

But since we have timestamp functionality already in some IPSO SOs, I would not object making it uniformly available as in this PR.

3203-1_1.xml
akeranen: I think the functionality of the value resource is not such that it would be measured, but rather that you can read exactly the value that was written there. Therefore there’s no uncertainty. 

The difference in range is a bit unfortunate. I think fixing that would require a new RR though. Since we are revving up the version, that’s possible option here.

3201-1_1.xml
dirtyha: Having multiple timestamp (and fractional timestamp) RR can be confusing. There is a danger that the new RR are not generic but specific to certain objects. I suggest to modify the current RR (timestamp and fractional timestamp) so that they fit this extended scenario. Maybe "Time when change was detected" is better since not all corresponding objects have value RR.

3203-1_1.xml
dirtyha: > I think the functionality of the value resource is not such that it would be measured, but rather that you can read exactly the value that was written there. Therefore there’s no uncertainty.

I'm ready to let this go and leave quality only in sensor objects. After all quality on sensor measurements is what matters to us.

3200-1_1.xml
dirtyha: > Any reasonable LwM2M server implementation should expose all information available from the clients. I guess the problem here is that using SenML time stamps was never properly defined in the LwM2M TS. That could be fixed in future versions.

I fully agree. We have done quite extensive evaluation of available LwM2M server implementations and the absence of senml timestamp in nb interface is unfortunately common. We need to be able to sell our LwM2M capable devices also "as-is" without our own cloud application. It would be sad to say to our customers that the device is not compatible with your LwM2M server, since it does not provide sensor measurement timestamps from senml messages. Instead we can say that your LwM2M server is not fully supported, but you have the workaround to read or observe timestamp resources in sensor object instances.

jaimejim (approved):

3203-1_1.xml
dirtyha: Yep, pushed in a change that removed quality resources from Digital Output (3203) and Positioner (3337).

akeranen: Thanks! And just to clarify: if you think there's a valid use case for the uncertainty here, I'm not against keeping them. Right now, with the given semantics of the objects, I just don't see one.

mkgillmore (approved):

PR #404, opened 2020-01-09
[Admin] OpenAIS merged into test

seanmcilroy29:
OpenAIS updated merged into test


jpradocueva:
This PR contains Objects that have been defined with "Ws" rather than "Joules".
We haven't received yet a confirmation from SenML Registry maintainers that "Ws" is accepted as a unit.
@akeranen clarification is requested. 
At the moment the PR fails validation until "Ws" is accepted as a unit in SenML.
See issue in https://github.com/core-wg/senml-more-units/issues/5

seanmcilroy29:
Ws unit value now validates with no errors using the latest schema in #409

PR #406, opened 2020-01-10
[Admin] Space removed from line 1 ObjID 10320

seanmcilroy29:
Space removed from line 1
Validated - Well formed


jpradocueva (changes requested):
ObjectID 10320 validates correctly now but there a whole resource defined as a comment towards the end of the file. This is not allowed in a file that is about to be registered.
See below the details in line 136.

image

seanmcilroy29:
The submitter will apply changes directly

PR #420, opened 2020-03-05
[ADMIN] Vaisala PR into Test

seanmcilroy29:
Vaisala branch
- [x] Validated with no errors
- [x] Validated well-formed

### DIFF
--- a/3338-1_1.xml
+++ b/3338-1_1.xml
@@ -1,0 +1,119 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<!-- BSD-3 Clause License
+
+Copyright 2019 Open Mobile Alliance. 
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions
+are met:
+
+1. Redistributions of source code must retain the above copyright
+notice, this list of conditions and the following disclaimer.
+2. Redistributions in binary form must reproduce the above copyright
+notice, this list of conditions and the following disclaimer in the
+documentation and/or other materials provided with the distribution.
+3. Neither the name of the copyright holder nor the names of its
+contributors may be used to endorse or promote products derived
+from this software without specific prior written permission.
+
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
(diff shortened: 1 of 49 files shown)


seanmcilroy29:
PR updated with OMA IPR Policy objects

 - [x] Validated with no errors
 - [x] Validated well-formed

3338-1_1.xml
```diff
@@ -0,0 +1,123 @@
...
+				<Operations>RW</Operations>
+				<MultipleInstances>Single</MultipleInstances>
+				<Mandatory>Optional</Mandatory>
+				<Type>String</Type>
+				<RangeEnumeration></RangeEnumeration>
+				<Units></Units>
+				<Description>The application type of the sensor or actuator as a string depending on the use case.</Description>
+			</Item>
+			<Item ID="5518">
+				<Name>Timestamp</Name>
```
akeranen, line 101: Is timestamp in buzzer intentional? "when measurement was performed" seems a bit odd here.

jpradocueva: @akeranen your comments are in a branch that was previously agreed by IPSO WG. 
The plan was to update this branch with the same changes as in Common. At that time there will be an option to update this content.

akeranen: Sounds good. Where should we make a note of that so that it's not forgotten?

jpradocueva: @akeranen @dirtyha    At the last IPSO call, it was agreed:

Before asking Harri  (Vaisala) to run his tool again to create versions 1.1 of IPSO objects Matt will provide a proposal on how to release Objects in this registry: block or single releases.
Once this is clear the staff will prepare the repository for that particular of release type. Once the repository is ready, we can ask to Harri to run his tool and create a new set of IPSO Objects. If we try to do it know we will have near twice the amount work when preparing the release of objects in this repository.

This release that I am referring to is that currently, all the versions of an object is in the same branch. But the group wants to remove versions from the filename, this implies that there will be a single version of an object in the same branch. Previous versions will be stored in  the "release branches".

jpradocueva:
This PR is closed as it will be superseded by a new PR that will create version 1.1 of the IPSO Obects.

PR #421, opened 2020-03-05
Unified description of application type RR

akeranen:


jpradocueva:
@akeranen there is something wrong in the PR and I cannot find out what is the problem.
When I validate your PR I am getting the errors listed below. But you are not doing modification of RRID 5751. 
Which baseline did you use to create your PR? Could you do a git PULL to ensure that you have the latest?

Error | Time | Description | ObjectID and/or ResourceID
-- | -- | -- | --
450 |   | Non existing reusable resource | 3200:5751
450 |   | Non existing reusable resource | 3202:5751
450 |   | Non existing reusable resource | 3300:5751
450 |   | Non existing reusable resource | 3302:5751

akeranen:
I used 7ad437bbfb7aad21ee1f704ee750d0e09e1848bf (latest in test branch).

jpradocueva:
In the LwM2M Editor Tool we removed the restriction of editing Reusable Resources. The implication is that now we have Reusable Resource in the Common.xml file (registered resources) and Reusable Resources in some Objects that doesn't match.
See below the update of changes in the tool. We will reverse this back again!

image

https://testdevtoolkit.openmobilealliance.org/OEditor/About

akeranen:
The key thing here is that RR description needs to be editable, but **not by the tool** since such changes should only be done very seldom and across all uses of the RR.

jpradocueva:
One of the validation check points is to verify that the description of the RR in the Object matches the description of the RR in the Common.xml file (the RR registration). If this description doesn't match the tool will through an error. 
 We are in the process to introduce this restriction in the tool (don't allow editing the RR description) and to re-instate the validation check point.

We need clear guidelines, do we want this validation and restrict the editing in the tool, YES/NO?

jpradocueva:
This PR cannot be merged until the problem with the RR description is resolved. The validation provides some errors that I still don't understand. So better to park it here until it is resolved.

jpradocueva:
@akeranen is this PR still needed?
If yes, could you please do a Pull from your branch to bring up to date your PR?

jpradocueva:
This PR is closed as it has been superseded by changes submitted by @akeranen .

PR #426, opened 2020-04-08
[Editorial] updates to RR Description in Common.xml

jpradocueva:
The aim of this PR is to correct any mistake in the Common.xml file. This file contains the Registry information for Reusable Resources. Once we get a correct baseline in the Common.xml file we an apply PRs to each Object.

Changes to RR:
* 5750
* 5601
* 5602

### DIFF
--- a/Common.xml
+++ b/Common.xml
@@ -1361,7 +1361,7 @@
 
         <Submitter>IPSO</Submitter>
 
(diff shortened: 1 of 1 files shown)


jpradocueva:
This PR is now covered by #441

PR #474, opened 2020-06-04
[Admin] Objects added to DDF

seanmcilroy29:
IPSO Objects added to DDF - 500 & 501


jpradocueva:
@seanmcilroy29 ok to merge into "BoD-22-June".

PR #475, opened 2020-06-05
[Admin] Aligning Object 9 resources with TS

seanmcilroy29:
Object 9 [Software Management] xml file is not aligned with the TS OMA-TS-LWM2M_SwMgmt-V1_0-20180301-A.

Resources 16, 17 and 18 have been added to the xml file to correspond with the Published TS


jpradocueva:
Having a second thought, we will need an agreement from DMSE on this. 
The addition of new Resources implies the creation of a new revision of the object.
The members will have to agree this change.

jpradocueva (commented):
Having a second thought, we will need an agreement from DMSE on this. 
The addition of new Resources implies the creation of a new revision of the object.
The members will have to agree this change.

jpradocueva:
DMSE agreed on Tuesday 16th that no changes should be applied to the Object. The alignment of the TS and Object will be done in a future release.

PR #476, opened 2020-06-05
[Admin] Object placeholders added

seanmcilroy29:
Object 500 & 501 placeholders added

Validated with no errors


jpradocueva (approved):
ok

PR #480, opened 2020-06-12
[Admin] TS link updates

seanmcilroy29:
TS link updated with Approved details

### DIFF
--- a/DDF.xml
+++ b/DDF.xml
@@ -219,7 +219,7 @@ These LwM2M Object Resources MUST only be changed by a LwM2M Bootstrap-Server or
         <DDF>LWM2M_LOCKWIPE-v1_0_2.xml</DDF>
         <Vorto></Vorto>
         <DDFLink>1</DDFLink>
-        <TS>http://www.openmobilealliance.org/release/LWM2M_LOCKWIPE/V1_0_1-20181003-A/OMA-TS-LWM2M_LockWipe-V1_0-20180615-A.pdf</TS>
+        <TS>http://www.openmobilealliance.org/release/LWM2M_LOCKWIPE/V1_0_2-202006016-A/OMA-TS-LWM2M_LockWipe-V1_0_1-20200616-A.pdf</TS>
         <TSLink>1</TSLink>
     </Item>
     <Item>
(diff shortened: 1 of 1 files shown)


DDF.xml
```diff
@@ -278,7 +278,7 @@ These LwM2M Object Resources MUST only be changed by a LwM2M Bootstrap-Server or
         <DDF>LWM2M_WLAN_connectivity4-v1_0_1.xml</DDF>
         <Vorto></Vorto>
         <DDFLink>1</DDFLink>
-        <TS>http://www.openmobilealliance.org/release/LWM2M_CONNMGMT/V1_0-20170314-A/OMA-TS-LWM2M_ConnMgmt-V1_0-20170314-A.pdf</TS>
+        <TS></TS>
         <TSLink>1</TSLink>
     </Item>
```
jpradocueva, line 283: If there is not a TS, then TSLink should be set to "0"

DDF.xml
```diff
@@ -219,7 +219,7 @@ These LwM2M Object Resources MUST only be changed by a LwM2M Bootstrap-Server or
         <DDF>LWM2M_LOCKWIPE-v1_0_2.xml</DDF>
         <Vorto></Vorto>
         <DDFLink>1</DDFLink>
-        <TS>http://www.openmobilealliance.org/release/LWM2M_LOCKWIPE/V1_0_1-20181003-A/OMA-TS-LWM2M_LockWipe-V1_0-20180615-A.pdf</TS>
+        <TS>http://www.openmobilealliance.org/release/LWM2M_LOCKWIPE/V1_0_2-20200616-A/OMA-TS-LWM2M_LockWipe-V1_0_1-20200616-A.pdf</TS>
```
jpradocueva, line 222: why is v1.0.2, it should be version v1.0.1, isn't?

jpradocueva (changes requested):
I have added comments that need to be address before merging this content into "bug-fix-DMSE-Objects" feature branch.

DDF.xml
seanmcilroy29: `V1_0_2-20200616-A` is the correct file name. 
The TS has been updated to `v1_0_1` to align with the changes but the file reference is aligned with the latest ERP/RRP version, which for this Enabler is v1.0.2

seanmcilroy29:
Validates with no errors with the exception of `Incorrect TS URL` due to files not being published yet

jpradocueva (approved):
can we merge this?

PR #482, opened 2020-06-17
Update with DMSE WG approved SUP files

jmudge:


LWM2M_Portfolio-v1_0_1.xml
```diff
@@ -15,17 +15,17 @@ NORMATIVE INFORMATION
 
   Information about this file can be found in the latest revision of
 
-    OMA-TS-LWM2M_PortfolioObj-V1_0
+    OMA-TS-LWM2M_PortfolioObj-V1_0_1
 
-  This is available at http://www.openmobilealliance.org/
+  This is available at https://www.omaspecworks.org/
```
jpradocueva: @jmudge Why do we have omaspecworks.org for Technical documents? We decided that we won't  change that location.

LWM2M_DevCapMgmt-v1_0_1.xml
```diff
@@ -15,17 +15,17 @@ NORMATIVE INFORMATION
...
+  This is available at https://www.omaspecworks.org/
 
   Send comments to https://github.com/OpenMobileAlliance/OMA_LwM2M_for_Developers/issues
 
 CHANGE HISTORY
 
 08032017 Status changed to Approved by TP, TP Ref # OMA-TP-2017-0021R01-INP_LWM2M_DevCapMgmt_1.0.2_for_Final_Approval_Notification
-16062020 Status changed to Approved by DMSE, Doc Ref # OMA-DM_SE-2020-0018-INP_LwM2M_DevCapMgmt_V1_0_3_RRP_for_Final_Approval
-	
+16062020 Status changed to Approved by DMSE, Doc Ref # OMA-DM&SE-2020-0018-INP_LWM2M_DevCapMgmt_V1_0_3_RRP_for_Final_Approval
```
jpradocueva, line 27: @jmudge in the past "&" created some issues. Shouldn't it be DMSE?

LWM2M_DevCapMgmt-v1_0_1.xml
```diff
@@ -15,17 +15,17 @@ NORMATIVE INFORMATION
 
   Information about this file can be found in the latest revision of
 
-    OMA-TS-LWM2M_DevCapMngt-V1_0
+    OMA-TS-LWM2M_DevCapMgmt-V1_0_3
 
-  This is available at http://www.openmobilealliance.org/
+  This is available at https://www.omaspecworks.org/
```
jpradocueva: omaspecworks.org in several places to refer to a TS location.

jpradocueva (changes requested):
I think that we need to discuss about some of the changes with regards to the location of the TS, the used of DM&SE.

LWM2M_Portfolio-v1_0_1.xml
jmudge: @jpradocueva the applicable template is https://github.com/OpenMobileAlliance/OMA_Document_Templates/blob/master/SUP/n.xml and this includes https://www.omaspecworks.org/ . I don't know when or whom initiated this change but it was over a year ago. If it is required to change the URL in these XML files then this template would also need to be changed.

Furthermore please note that http://www.openmobilealliance.org/ automatically redirects to https://www.omaspecworks.org/ . People have contacted the Helpdesk complaining that they are unwilling to use the auto diverted URL https://www.omaspecworks.org/ as they feared that the site http://www.openmobilealliance.org/ had been hijacked to redirect to an illegitimate site.

Please confirm if you wish the URL in these XML files and also in the template to be changed.

LWM2M_DevCapMgmt-v1_0_1.xml
jmudge: Please see the following screenshot from the portal:

DM SE_document_ref

The current reference in LWM2M_DevCapMgmt-v1_0_1.xml seems to be correct.

From http://www.w3.org/TR/REC-xml/#syntax "... The ampersand character (&) and the left angle bracket (<) MUST NOT appear in their literal form, except when used as markup delimiters, or within a **comment**, a processing instruction, or a CDATA section. ..."

LWM2M_DevCapMgmt-v1_0_1.xml
jmudge: Please see the previous related reply.

Please confirm if you wish the URL in these XML files and also in the template https://github.com/OpenMobileAlliance/OMA_Document_Templates/blob/master/SUP/n.xml to be changed.

seanmcilroy29 (approved):

LWM2M_DevCapMgmt-v1_0_1.xml
jpradocueva: My view is that the link should point to: http://www.openmobilealliance.org/release/ 
This is generic enough and it can be setup in the template by default. If this is acceptable we will have to change the templates and rectify this pointer in all the objects that we try to rectify.

LWM2M_DevCapMgmt-v1_0_1.xml
jpradocueva: With regards to &. Ok.

LWM2M_DevCapMgmt-v1_0_1.xml
jmudge: @jpradocueva as requested the public URLs have been updated.

jpradocueva (changes requested):
Failed validation.
The used of (..) instead of (-) should be restricted to RangeEnurmeration only or to fields that don't raise errors. By introducing changes in the description of  Reusable Resources 4010 and 4013. The tool detect changes

jpradocueva:
The validation shows that it cannot find the TSs but this ok as the TS won't be added to the FTP after the Board ratification on June  22nd.

PR #483, opened 2020-06-18
[Admin] Approved TS version updates

seanmcilroy29:
DMSE - Approved TS version updated and URL updated to https://www.omaspecworks.org/

### DIFF
--- a/LWM2M_LOCKWIPE-v1_0_2.xml
+++ b/LWM2M_LOCKWIPE-v1_0_2.xml
@@ -17,7 +17,7 @@ NORMATIVE INFORMATION
 
     OMA-TS-LightweightM2M-V1_0_1
 
-  This is available at https://www.omaspecworks.org/
+  This is available at http://www.openmobilealliance.org/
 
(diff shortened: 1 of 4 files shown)


jpradocueva (changes requested):
I fail to understand why the location has been changed to https://www.omaspecworks.org/

jpradocueva:
From my point of view the link should point to: http://www.openmobilealliance.org/release/ as a generic link that can be added to the template.

seanmcilroy29:
The updates are in line with the SUP template you updated https://github.com/OpenMobileAlliance/OMA_Document_Templates/blob/master/SUP/n.xml.. where the template specifies - https://www.omaspecworks.org/

I'm not aware that the generic link has ever been http://www.openmobilealliance.org/release/, I believe it's always been http://www.openmobilealliance.org.

I'll make the updates to these files, as you've proposed and then I'll update the TS templates

jpradocueva (approved):
Ready to merge.

PR #501, opened 2020-07-09
[ADMIN] creation of rel 3 with ONLY new objects

jpradocueva:
new pull request template


jpradocueva:
The IPSO WG decided to follow a different approach. Therefore, the PR can be closed.
A 'snapshots' folder will be added that will contain all the objects with the version in the file name. The root branch will contain the latest available version of each object and without version in the file name.

Issue #502, opened 2020-07-09
[Sierra Wireless] Object Registration  `DeviceExt` 


sbernard31:
### 1. Choose the type of registration: ###
  - [X] ```Object``` 
  - [ ] ```Reusable Resource```
  
### 2. Choose the [Object Class](http://www.openmobilealliance.org/wp/OMNA/LwM2M/LwM2MRegistry.html#registry-table): ###
  - [ ] 3rd party Standards or Alliances
  - [X] Private companies
  - [ ] Private individuals (not allowed Reusable Resources registration)
  
## 3. Confirm Software License
  - [X] Tick this box confirming that your submission has included a [software license](http://devtoolkit.openmobilealliance.org/OEditor/License).
  

Concept            | Description
:----------------|:------------------------------
Number of ```Object(s)``` or ```Reusable Resource(s)``` to register  | 1
Name of SDO, Company or individual    | Sierra Wireless
Object Name (as defined in the .xml file) |  DeviceExt (but we are open to better proposition)
Short Object description, to be used in the pubic registry (max 134 characteres) | This object extends Device(/3) Object. It mainly provides features allowing better monitor of Reboot and Reset Factory operation.

## Some context :

This contribution is motivated by the issue raised at : https://github.com/OpenMobileAlliance/OMA_LwM2M_for_Developers/issues/280#issuecomment-363483045

A first draft is visible at : https://rentry.co/edzod

seanmcilroy29:
@sbernard31 Thank you for your registration.

**Required Object Updates**
- [ ] Object Version 1.0
- [ ] Object ID - 10371
- [ ] The object URN allocation for this object will be reserved as urn:oma:lwm2m:x:10371
- [ ] Update range enumeration from ``-`` to ``..``

**Steps** A new branch,  ```Sierra Wireless``` has been created in the LwM2M registry for the WG to review your object - Please submit a Pull Request with the content of your object against this featured branch: https://github.com/OpenMobileAlliance/lwm2m-registry/tree/Sierra_Wireless

Steps | Actions
-- | --
1 | Resolve observations
2 | Submit Pull Request to the nominated branch with your object
3 | Staff validates PR
4 | IPSO Working Group discuss the submitted Object registration
5 | Any issues will require the submitter to update the PR accordingly
6 | Staff validates object again after submitters updates
7 | IPSO Working Group Chair will merge the object
8 | Staff publish the Object to the OMNA LwM2M Registry

[...]

PR #504, opened 2020-07-16
[Sierra_Wireless]  Reboot Status - Object 10371

sbernard31:
This is triggered by #502.

This aims to resolves : https://github.com/OpenMobileAlliance/OMA_LwM2M_for_Developers/issues/280.

IMHO, this should be part of the LWM2M specification (directly added to Device object) and should not require an extension.

But for LWM2M 1.0.x, it's too late and an new object is needed to resolve this issue.
For next version of the specification, maybe this should be considered as an addition ? (making this object obsolete in the future)

### DIFF
--- a/10371.xml
+++ b/10371.xml
@@ -1,0 +1,85 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!--Copyright 2020 Sierra Wireless. 
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions
+are met:
+
+1. Redistributions of source code must retain the above copyright
+notice, this list of conditions and the following disclaimer.
+2. Redistributions in binary form must reproduce the above copyright
+notice, this list of conditions and the following disclaimer in the
+documentation and/or other materials provided with the distribution.
+3. Neither the name of the copyright holder nor the names of its
+contributors may be used to endorse or promote products derived
+from this software without specific prior written permission.
+
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+"AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+POSSIBILITY OF SUCH DAMAGE.
+
+-->
+<LWM2M xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="http://www.openmobilealliance.org/tech/profiles/LWM2M.xsd">
+	<Object ObjectType="MODefinition">
+		<Name>Reboot Status</Name>
+		<Description1><![CDATA[This object provides features allowing to monitor Reboot and Reset Factory operation from LWM2M Device object.]]></Description1>
+		<ObjectID>10371</ObjectID>
+		<ObjectURN>urn:oma:lwm2m:x:10371</ObjectURN>
+		<LWM2MVersion>1.0</LWM2MVersion>
+		<ObjectVersion>1.0</ObjectVersion>
+		<MultipleInstances>Single</MultipleInstances>
+		<Mandatory>Optional</Mandatory>
+		<Resources>
+			<Item ID="0">
+				<Name>Reboot State</Name>
+				<Operations>R</Operations>
+				<MultipleInstances>Single</MultipleInstances>
+				<Mandatory>Mandatory</Mandatory>
+				<Type>Integer</Type>
+				<RangeEnumeration>0..2</RangeEnumeration>
+				<Units></Units>
+				<Description><![CDATA[ 	Containing the state of reboot operation. At device start-up the value should be 0 (booted). Once a reboot process is scheduled (e.g. exec on /3/0/4), this value must be set to 1 (rebooting), it is adviced to do that before to send response of Reboot resource (/3/0/4) to avoid any race conditions. If rebooting failed for any reason value must be set to 2(reboot error). If reboot succeeds, at device start up, reboot state should be set to 0 (booted) as always.]]></Description>
+			</Item>
+			<Item ID="1">
+				<Name>Reboot Error Message</Name>
+				<Operations>R</Operations>
+				<MultipleInstances>Single</MultipleInstances>
+				<Mandatory>Mandatory</Mandatory>
+				<Type>String</Type>
+				<RangeEnumeration></RangeEnumeration>
+				<Units></Units>
+				<Description><![CDATA[Containing the error message of the last reboot failure. This value MUST be set to an empty string on exec on /3/0/4, ideally before to send the response for Reboot resource (3/0/4) to avoid any race conditions]]></Description>
+			</Item>
+			<Item ID="2">
+				<Name>Reset Factory State</Name>
+				<Operations>R</Operations>
+				<MultipleInstances>Single</MultipleInstances>
+				<Mandatory>Optional</Mandatory>
+				<Type>Integer</Type>
+				<RangeEnumeration>0..3</RangeEnumeration>
+				<Units></Units>
+				<Description><![CDATA[ 		Containing the state of reset factory operation. At start-up the value should be 0 (initial-state). Once the reset factory process is scheduled (exec on /3/O/5), this value must be set to 1 (resetting), it is advised to do that before to send response of Reset Factory resource (/3/0/5) to avoid any race conditions. If reset fails for any reason value must be set to 2 (reset error). If reset succeeds, state should be set to 3 (reset). If the reset factory need a reboot, reboot state(/10371/0/0) resource SHOULD also be updated accordingly. If reset factory need a reboot, in case of successful reset, the value MUST be set to 3 (reset) at device start-up. If it helps implementation, this is tolerated to have this resource set to 3 (reset) at first device start-up.]]></Description>
+			</Item>
+			<Item ID="3">
+				<Name>Reset Factory Error Message</Name>
+				<Operations>R</Operations>
+				<MultipleInstances>Single</MultipleInstances>
+				<Mandatory>Optional</Mandatory>
+				<Type>String</Type>
+				<RangeEnumeration></RangeEnumeration>
+				<Units></Units>
+				<Description><![CDATA[Containing the error message of the last reset factory failure. This value MUST be set to an empty string on exec on /3/0/5, ideally before to send the response for Reset Factory resource (3/0/5) to avoid any race conditions]]></Description>
+			</Item>
+		</Resources>
+		<Description2><![CDATA[This LWM2M object provides features to monitor Reboot(/3/0/4) and Reset Factory(/3/0/5) operation available on LWM2M Device (3) Object, allowing to know if a reboot or a reset factory is a success or a failure and why it failed. This is strongly recommended to use separated response for those resources (/3/0/4 and /3/0/5) and wait for response ACK before to execute/schedule the operation (reboot or reset factory). This prevents LWM2M server to consider the request as a failure in case response is lost. This situation could be annoying as server could decide to resent the request later which could bring to unexpected reboot or reset factory.]]></Description2>
+	</Object>
+</LWM2M>
--- a/DDF.xml
+++ b/DDF.xml
@@ -3982,7 +3982,7 @@ These LwM2M Object Resources MUST only be changed by a LwM2M Bootstrap-Server or
     <Item>
         <ObjectID>10371</ObjectID>
         <URN>urn:oma:lwm2m:x:10371</URN>
-        <Name>DeviceExt</Name>
+        <Name>Reboot Status</Name>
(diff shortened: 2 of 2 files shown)


10371.xml
```diff
@@ -0,0 +1,85 @@
...
+CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+POSSIBILITY OF SUCH DAMAGE.
+
+-->
+<LWM2M xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="http://www.openmobilealliance.org/tech/profiles/LWM2M.xsd">
+	<Object ObjectType="MODefinition">
+		<Name>DeviceExt</Name>
+		<Description1><![CDATA[This object extends Device(/3) Object. It mainly provides features allowing better monitor of Reboot and Reset Factory operation.]]></Description1>
```
sbernard31: I use a short description here as it is adviced to use 134 character max.

(see https://github.com/OpenMobileAlliance/lwm2m-registry/wiki/Guidelines#description)

10371.xml
```diff
@@ -0,0 +1,85 @@
...
+				<Type>String</Type>
+				<RangeEnumeration></RangeEnumeration>
+				<Units></Units>
+				<Description><![CDATA[Containing the error message of the last reset factory failure. This value MUST be set to an empty string on exec on /3/0/5, ideally before to send the response for Reset Factory resource (3/0/5) to avoid any race conditions]]></Description>
+			</Item>
+		</Resources>
+		<Description2><![CDATA[This LWM2M object extends Device(3) from LWM2M core specification. It mainly provides features allowing better monitor of Reboot(/3/0/4) and Reset Factory(/3/0/5) operation available on LWM2M Device (3) Object. Without this extension this is really hard to know if a reboot or a reset factory is a success or a failure and why it failed. This is also strongly recommended to use separated response for those resources (/3/0/4 and /3/0/5) and wait for response ACK before to execute/schedule the operation (reboot or reset factory). This prevents LWM2M server to consider the request as a failure in case response is lost. This situation could be annoying as server could decide to resent the request later which could bring to unexpected reboot or reset factory.]]></Description2>
```
sbernard31: I use `Description2` for a more complete description. (I don't know if this is the idea)

jpradocueva: Description 2 is used for a variety of situations. Therefore, it is ok to add further details if needed.

10371.xml
akeranen: If this "extends" the device object, should these be considered for a new version of the device object itself? If that's not good option, perhaps better to say "provides added functionality beyond.." or similar. And at large, maybe we (DMSE?) 
should think what kind of paradigm and design pattern we really want for this kind of extensions to the management objects.

10371.xml
akeranen: > Without this extension this is really hard to know if a reboot or a reset factory is a success [...]

This kind of description sounds weird in an object. Makes sense as part of the discussion here when deciding if/how this should be added, but not as long-term description in an object.

10371.xml
```diff
@@ -0,0 +1,85 @@
...
+LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+POSSIBILITY OF SUCH DAMAGE.
+
+-->
+<LWM2M xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="http://www.openmobilealliance.org/tech/profiles/LWM2M.xsd">
+	<Object ObjectType="MODefinition">
+		<Name>DeviceExt</Name>
```
akeranen: I'd recommend full words for the name. For example "Extended Device Management".

dnav: Maybe a more specific name ? Like "Reboot Status" as this Object's Resources are dealing only with the reboot and factory reset states.

10371.xml
```diff
@@ -0,0 +1,85 @@
...
+				<Type>String</Type>
+				<RangeEnumeration></RangeEnumeration>
+				<Units></Units>
+				<Description><![CDATA[Containing the error message of the last reboot failure. This value MUST be set to an empty string on exec on /3/0/4, ideally before to send the response for Reboot resource (3/0/4) to avoid any race conditions]]></Description>
+			</Item>
+			<Item ID="2">
+				<Name>Reset Factory State</Name>
+				<Operations>R</Operations>
+				<MultipleInstances>Single</MultipleInstances>
+				<Mandatory>Mandatory</Mandatory>
```
dnav: The "Reboot" feature is mandatory to support for LwM2M Device but "Factory Reset" is optional. Thus I would suggest to make this Resource Optional.

10371.xml
```diff
@@ -0,0 +1,85 @@
...
+				<Type>Integer</Type>
+				<RangeEnumeration>0..3</RangeEnumeration>
+				<Units></Units>
+				<Description><![CDATA[ 		Containing the state of reset factory operation. At start-up the value should be 0 (initial-state). Once the reset factory process is scheduled (exec on /3/O/5), this value must be set to 1 (resetting), it is advised to do that before to send response of Reset Factory resource (/3/0/5) to avoid any race conditions. If reset fails for any reason value must be set to 2 (reset error). If reset succeeds, state should be set to 3 (reset). If the reset factory need a reboot, reboot state(/10371/0/0) resource SHOULD also be updated accordingly. If reset factory need a reboot, in case of successful reset, the value MUST be set to 3 (reset) at device start-up. If it helps implementation, this is tolerated to have this resource set to 3 (reset) at first device start-up.]]></Description>
+			</Item>
+			<Item ID="3">
+				<Name>Reset Factory Error Message</Name>
+				<Operations>R</Operations>
+				<MultipleInstances>Single</MultipleInstances>
+				<Mandatory>Mandatory</Mandatory>
```
dnav: The "Reboot" feature is mandatory to support for LwM2M Device but "Factory Reset" is optional. Thus I would suggest to make this Resource Optional.

10371.xml
sbernard31: Definitely a new version of Device Object would be better.
This Object just aims to fix that it seems to me as an LWM2M core issue that we discuss in https://github.com/OpenMobileAlliance/OMA_LwM2M_for_Developers/issues/280#issuecomment-363483045

But we need this new version for LWM2M 1.0.x and I don't know if this is doable ? 
If this is possible then we don't really need this new object.

10371.xml
sbernard31: If we keep this object (https://github.com/OpenMobileAlliance/lwm2m-registry/pull/504/files#r463470429) I will change it.

10371.xml
sbernard31: If we keep this object (https://github.com/OpenMobileAlliance/lwm2m-registry/pull/504/files#r463470429) I will change it.

10371.xml
sbernard31: The idea was to explain rational of this object.
I thought that it could help people to decide if they need to implement it or not.

10371.xml
sbernard31: I'm not strongly attached to the name, so no problem to change it.
Maybe `Reboot Status` or `Device Status` could be OK.
Let me know which name sounds more appropriate to you ?
Then If we decide to keep this object (https://github.com/OpenMobileAlliance/lwm2m-registry/pull/504/files#r463470429) I will change it.

sbernard31:
(Just to let you know I will be unavailable the next 2 weeks)

10371.xml
akeranen: For that I’d recommend wording like “This object provides additional information on whether xxx succeeded...” etc. “really hard to know” sounds very subjective.

seanmcilroy29:
@sbernard31 - Can you confirm if all the submitted observations have been resolved?

sbernard31:
@seanmcilroy29, I didn't integrate any changes, because : 

https://github.com/OpenMobileAlliance/lwm2m-registry/pull/504#discussion_r463441342 raises the question about moving this in Device Object directly.

I also think this would be better (https://github.com/OpenMobileAlliance/lwm2m-registry/pull/504#discussion_r463470429). but this means we need to create a v1.2 for Device object.
The issue is that we need this object version available for lwm2m 1.0 and Device object v1.1 already target lwm2m 1.1.

My guess is that this DeviceExt object is needed at least for lwm2m 1.0, but I prefer to wait for OMA feedback/confirmation before to resolve observation.

sbernard31:
@seanmcilroy29 , @dnav , @akeranen any news about my concern :point_up:  (https://github.com/OpenMobileAlliance/lwm2m-registry/pull/504#issuecomment-677443802)

akeranen:
This is an interesting question and shows some of the problems of supposedly-linear versioning. If we had features instead of versions of the protocol, you could use all versions of Device object with all versions of the protocol and just not use the features not supported. Something to think about for v2.0. Meanwhile, if you really need these to work with 1.0, new object seems like the best approach.

sbernard31:
@akeranen, thx for the answer. I will apply requested changes to this PR now.

sbernard31:
I hope I take all the comments into account.
I did it in a separated commit to make review of new changes easier, but I can squash all commit in one if needed.

seanmcilroy29:
@sbernard31 could you squash all your commits into this PR.
Tomorrow IPSO will be holding a Conf Call and this Object will be submitted for Approval

sbernard31:
@seanmcilroy29  done. :)

jpradocueva:
Since this Pull Request was created the "test" branch has undergone major changes.
Once there is an agreement with this Object. The maintainer will move the Object directly to the new "test" branch.

jpradocueva:
The name of the ObjectID was updated in the DDF.xml to match the name of the Object "Reboot Status". 
Now it passes validation.

mojanm:
Having read through https://github.com/OpenMobileAlliance/OMA_LwM2M_for_Developers/issues/280 that has triggered this PR, the proposed object certainly seems to address the issue and looks good. The only concern I have is doing exec commands on resources of device object suddenly triggers changes to this object without any relationship being established between the two objects which in my view is something of a departure for LwM2M? I fully understand the desire to avoid making significant changes to device object at this stage, but maybe it would be possible to add executable resources for Factory reset and Reboot to this object so that we have a self contained reboot/factory reset object and for the clients that support this object the server can control reboot and factory reset via this object entirely? Factory reset is an optional resource in the device object, and its duplication can be avoided. However, reboot is mandatory which means we have a duplicate resource in two objects, but to me that we can live with.

10371.xml
```diff
@@ -0,0 +1,85 @@
...
+LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+POSSIBILITY OF SUCH DAMAGE.
+
+-->
+<LWM2M xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="http://www.openmobilealliance.org/tech/profiles/LWM2M.xsd">
+	<Object ObjectType="MODefinition">
+		<Name>Reboot Status</Name>
```
jaimejim, line 33: "Reboot Status" sounds good to me.

10371.xml
```diff
@@ -0,0 +1,85 @@
...
+				<Operations>R</Operations>
+				<MultipleInstances>Single</MultipleInstances>
+				<Mandatory>Optional</Mandatory>
+				<Type>Integer</Type>
+				<RangeEnumeration>0..3</RangeEnumeration>
+				<Units></Units>
+				<Description><![CDATA[ 		Containing the state of reset factory operation. At start-up the value should be 0 (initial-state). Once the reset factory process is scheduled (exec on /3/O/5), this value must be set to 1 (resetting), it is advised to do that before to send response of Reset Factory resource (/3/0/5) to avoid any race conditions. If reset fails for any reason value must be set to 2 (reset error). If reset succeeds, state should be set to 3 (reset). If the reset factory need a reboot, reboot state(/10371/0/0) resource SHOULD also be updated accordingly. If reset factory need a reboot, in case of successful reset, the value MUST be set to 3 (reset) at device start-up. If it helps implementation, this is tolerated to have this resource set to 3 (reset) at first device start-up.]]></Description>
+			</Item>
+			<Item ID="3">
+				<Name>Reset Factory Error Message</Name>
```
jaimejim, line 73: Error message is an unformatted string, with a dump of the error. I guess in the long run we might want to have something more specific about how the error is formatted. That's for OMA to specify, not on this Object in particular.

10371.xml
```diff
@@ -0,0 +1,85 @@
...
+				<Operations>R</Operations>
+				<MultipleInstances>Single</MultipleInstances>
+				<Mandatory>Optional</Mandatory>
+				<Type>String</Type>
+				<RangeEnumeration></RangeEnumeration>
+				<Units></Units>
+				<Description><![CDATA[Containing the error message of the last reset factory failure. This value MUST be set to an empty string on exec on /3/0/5, ideally before to send the response for Reset Factory resource (3/0/5) to avoid any race conditions]]></Description>
+			</Item>
+		</Resources>
+		<Description2><![CDATA[This LWM2M object provides features to monitor Reboot(/3/0/4) and Reset Factory(/3/0/5) operation available on LWM2M Device (3) Object, allowing to know if a reboot or a reset factory is a success or a failure and why it failed. This is strongly recommended to use separated response for those resources (/3/0/4 and /3/0/5) and wait for response ACK before to execute/schedule the operation (reboot or reset factory). This prevents LWM2M server to consider the request as a failure in case response is lost. This situation could be annoying as server could decide to resent the request later which could bring to unexpected reboot or reset factory.]]></Description2>
```
jaimejim, line 83: As others have mentioned, instead of extending, it would be good to directly update /3 with these resources. Maybe DM can have a look and add it to the spec at some point?

jaimejim (approved):
I provided some comments on the review but all in all the current version is valid. I also checked with the validator and it passes validation (I forgot that we have that automated nowadays :D ) 

I also have the same comment as @mojanm that we are adding extension objects that trigger executable resources on DM objects might have unintended consequences. It would be better to extend DM objects instead.

sbernard31:
As I said before, I agree that this makes probably more sense to put this is a device object but it sounds this is not possible to get it compatible for LWM2M 1.0 (see https://github.com/OpenMobileAlliance/lwm2m-registry/pull/504#issuecomment-677443802 and https://github.com/OpenMobileAlliance/lwm2m-registry/pull/504#issuecomment-682666160)

About, duplicating executable resources from Device to this Object, I can not see the benefits of this.
I don't even sure what would that means concretely :thinking:. 
Reboot status would only be available when exec resource from this object is used but exec on device object will not change status of this object ? (meaning this is more complicate for devices to implement)
or both resource will do exactly the same thing and so I do not understand the what this is better.

By the way, if we do that the name and the documentation of this object is no more adapted.

mkgillmore:
Approved via R&A pending validation

PR #507, opened 2020-07-18
[Admin] IPSO Objects 502 & 503 added

seanmcilroy29:
IPSO Approved Objects 502 & 503 added

Feature branch validated with no errors


jpradocueva (approved):
ok

PR #510, opened 2020-07-22
[Admin] Permanent Document file path added

seanmcilroy29:
Permanent Document file path added to 502 & 503


jpradocueva (approved):
Ok to merge into test

PR #512, opened 2020-07-30
[CONR][B] LWM2M_DevCapMgmt-v1_0_1.xml, Resource 6 (opDisable) multiplicity is incorrect 

mojanm:
Resource 6 in OMA-TS-LWM2M_DevCapMgmt-V1_0_2-20170308-A is defined as Single while xml file shows it as Multiple which cannot be the case for an executable resource as per Core spec. Given Resource 6 is a mandatory resource, fixing this conflict will result in creating version 2.0 of the object.

### DIFF
--- a/LWM2M_DevCapMgmt-v1_0_1.xml
+++ b/LWM2M_DevCapMgmt-v1_0_1.xml
@@ -67,7 +67,8 @@ LEGAL DISCLAIMER
 		<Name>DevCapMgmt</Name>
 		<Description1><![CDATA[This LWM2M Object is dedicated to manage the device capabilities of a device e.g. sensors, communication, etc.]]></Description1>
 		<ObjectID>15</ObjectID>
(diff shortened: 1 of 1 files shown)


seanmcilroy29:
@mojanm  - Could you apply your PR against the DevCapMgmt feature branch -https://github.com/OpenMobileAlliance/lwm2m-registry/tree/DevCapMgmt

PR's aren't applied against the test branch directly

Also, this should be discussed with DMSE WG

seanmcilroy29:
@mojanm - Are you proposing v1.0 is deprecated by v2.0

mkgillmore:
@ssniot @hannestschofenig  This pull request needs to be handled in DM because the TS is different than the object

jpradocueva:
The content of this PR will be discussed in DMSE. Therefore, we are closing the PR here. See above.

PR #517, opened 2020-08-11
[ADMIN] The changes in this PR should be applied to the script

jpradocueva:
This PR contains changes that should be implemented in Matt's script.
In this way, future updates made by the script won't reintroduce these problems again.


mkgillmore:
Group agrees to merge with fixing object 20V1_0.xml change to 20-1_0.xml and removing ddf.xml and common.xml from the version_history folder. The schema files were agreed also to be removed.

PR #522, opened 2020-09-04
[ADMIN] merge latest changes from "test" into "prod"

jpradocueva:
This PR introduces in production the latest changes in the LwM2M-Registry:
* In the root of the "master" or "prod" branch all the Objects are identified now by its own `ObjectID` rather than the `Object Name`
* The root of the branch contains ONLY the latest versions or revisions of the Objects
* Previous versions of the Objects are stored in a new folder called `version_history`, which contains ALL the Objects in the repository. The name of the Objects in this folder is called: `ObjectID-x_y_z`, where x, y and z are the version and service indicator numbers.


jpradocueva (commented):
It validates ok

PR #531, opened 2020-10-08
uCIFI fix units in object 3430

uCIFI-Technical-Workgroup:
Fixing units lat, lon and degrees in object 3430


seanmcilroy29:
This PR is pushing the PROD feature branch into the uCIFI which is not permitted.
All PRs must be submitted only **against** the uCIFI branch

PR #533, opened 2020-10-13
[ADMIN] Update Object 3 with DMSE changes

jpradocueva:
DMSE update object 3 with a clarification on the kilobytes definition, 1 kilobyte corresponds to 1000 bytes.


jpradocueva:
I am closing this PR due to problems with the CR and LF.

Issue #541, opened 2020-11-05
Smart Quotes Object Errors


seanmcilroy29:
Error | Time | Description | ObjectID and/or ResourceID | Error update
-- | -- | -- | -- | -- 
420 |   | Invalid XML characters | 1:4 |  https://github.com/OpenMobileAlliance/objects-lwm2m/pull/156
420 |   | Invalid XML characters | 1:6 | https://github.com/OpenMobileAlliance/objects-lwm2m/pull/156
420 |   | Invalid XML characters | 1:8 |  https://github.com/OpenMobileAlliance/objects-lwm2m/pull/156
420 |   | Invalid XML characters | 8:0 |  Smart quotes
420 |   | Invalid XML characters | 9:5 |  https://github.com/OpenMobileAlliance/SwMgmt to be updated
420 |   | Invalid XML characters | 9:6 |  https://github.com/OpenMobileAlliance/SwMgmt to be updated
420 |   | Invalid XML characters | 9:9 |  https://github.com/OpenMobileAlliance/SwMgmt to be updated
420 |   | Invalid XML characters | 9:13 |  https://github.com/OpenMobileAlliance/SwMgmt to be updated
420 |   | Invalid XML characters | 10:2 |  Star (*) Error
705 |   | Invalid XML characters | 11 |  https://github.com/OpenMobileAlliance/objects-lwm2m/pull/156
420 |   | Invalid XML characters | 11:7 |   https://github.com/OpenMobileAlliance/objects-lwm2m/pull/156
420 |   | Invalid XML characters | 11:8 |  Hyphen (-) error
420 |   | Invalid XML characters | 12:13 |  https://github.com/OpenMobileAlliance/objects-lwm2m/pull/156
420 |   | Invalid XML characters | 12:17 |  Hyphen (-) error
420 |   | Invalid XML characters | 13:6 |  https://github.com/OpenMobileAlliance/objects-lwm2m/pull/156
420 |   | Invalid XML characters | 13:7 |  https://github.com/OpenMobileAlliance/objects-lwm2m/pull/156
420 |   | Invalid XML characters | 15:3 |  https://github.com/OpenMobileAlliance/DevCapMgmt/pull/31
420 |   | Invalid XML characters | 15:4 |  https://github.com/OpenMobileAlliance/DevCapMgmt/pull/31
420 |   | Invalid XML characters | 15:7 |  https://github.com/OpenMobileAlliance/DevCapMgmt/pull/31
[...]

PR #542, opened 2020-11-05
[Bug] Remove smart quotes from published Objects

jpradocueva:
This PR addresses the problems identified by #541 on the following Objects:
* 2051
* 2053
* 10245
Still pending: (still under investigation)
* 10256 (problem with "--") The tool is identifying "-" as an incorrect character.
* 10309 (problem with "-") The tool is identifying "-" as an incorrect character.

The errors related to DMSE Objects will be resolved in a separately PR.

### DIFF
--- a/10245.xml
+++ b/10245.xml
@@ -169,7 +169,7 @@ POSSIBILITY OF SUCH DAMAGE.
                 <Type>Integer</Type>
                 <RangeEnumeration>-24..24</RangeEnumeration>
                 <Units>dB</Units>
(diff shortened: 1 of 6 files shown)


jpradocueva:
@akeranen if I recall correctly, you identified some problems when parsing Objects with smart quotes.
What about "-" does your tool also complains about this character?
There are two objects: where the tool identifies "-" as an incorrect character. 
* 10256 (problem with "--") 
* 10309 (problem with "-") 

Is your tool having problems with this character?

seanmcilroy29 (approved):

jmudge (commented):
DDF.xml ObjectID 10309 Description field needs to be updated to match 10309.xml
PR https://github.com/OpenMobileAlliance/lwm2m-registry/pull/544 has been created to do this

mkgillmore:
Group agrees with these changes 11/10/2020.  These changes also need to be made in the version_history objects

PR #549, opened 2020-11-13
[ADMIN] LwM2M DMSE Version-History update 

seanmcilroy29:
LwM2M DMSE Object Version-History update 

Validated with no additional errors with the exception of the known non-core/Non-OMA objects.


jpradocueva (approved):
ok, to merge

PR #550, opened 2020-11-13
[ADMIN] DEV Version-History updated 

seanmcilroy29:
DEV Version-History updated
Validated with no additional errors with the exception of the known OMA Objects.


jpradocueva (changes requested):
This PR is not correct. It changes objects that should not be changed and it contains changes that are incorrect.
The PR should be closed.

jpradocueva:
I am closing the PR just in case ....

PR #560, opened 2020-12-14
[Admin] Updating Approval date and Version

seanmcilroy29:
Updating Approval date, DDF TS URL and Version

Passed validation with the exception of `Incorrect TS URL`, which will be cleared when the TS is uploaded

### DIFF
--- a/15.xml
+++ b/15.xml
@@ -4,7 +4,7 @@
 FILE INFORMATION
 
 OMA Permanent Document
-  File: OMA-SUP-XML_15-V1_0_2-20201124-D
+  File: OMA-SUP-XML_15-V2_0-20201208-A
   Path: http://www.openmobilealliance.org/release/LWM2M_DevCapMgmt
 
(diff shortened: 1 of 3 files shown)


jmudge (approved):

jpradocueva (changes requested):
@seanmcilroy29 same changes should be done in the Object 15-2_0.xml inside of the "version_history" folder.

seanmcilroy29:
@jpradocueva - The version_history update for v2.0 I left to see if the staff could have access to Matt's tool and apply the update that way. 

I've applied the update manually, but I'll continue to make a request to Matt for staff access to the tool.

seanmcilroy29:
TS Place holder added - Validates with no errors

jpradocueva (approved):
Ok to merge into "test_15" branch, but we have to wait for BoD Approval before moving it to "development" branch.

PR #562, opened 2020-12-30
[ADMIN] Resolve mismatch between Descriptions

jpradocueva:
It appears that the validation tool is not checking if the value of the element "Description" in the DDF is the same as the value of the "Description1" in the Object. The Validation tool is being updated to perform this check.


jpradocueva:
@seanmcilroy29 @jmudge there is something occur with this PR. I believe that my working drive copy was not correct. I created the new branch "dev_Description1" from "dev" and I am getting, in the PR, commits that I don't understand. I will close the PR and branch and start again.

PR #574, opened 2021-01-19
[ADMIN] New revisions of objects 9 and 14

jmudge:
Objects 9 and 14 revised to match the template.


seanmcilroy29 (approved):

jpradocueva (approved):
Thank you John, good catch!

PR #576, opened 2021-01-28
[ADMIN] Publish January releases

jmudge:


seanmcilroy29 (approved):

jpradocueva:
@seanmcilroy29 @jmudge personally I prefer to merge finalize content in the "test" branch. In this case, we know that the Error 800 is still pending in the IPSO objects.
Also, we have identified that the tool was not rendering correctly some of the angular brackets, at that time we change the descriptions to use  e.g. "&lt" and now that we are resolving that problem the tool is not rendering correctly "&lt". This topic should be investigated and finalized before merge content into "test".

seanmcilroy29:
Dev Branch now validates with no errors

jpradocueva (approved):
Ok to merge.

Issue #567, opened 2021-01-08
Object 10243, Voltage as type String, should be float


kdhaka:
Single Phase Power Meter Tension (Voltage) type is String. It should be Float:
https://github.com/OpenMobileAlliance/lwm2m-registry/blob/prod/10243.xml#L97

3 Phase Power Meter Tension type is correct:
https://github.com/OpenMobileAlliance/lwm2m-registry/blob/prod/10242.xml#L97

jpradocueva:
Prepare a new branch where the DDF will be updated to v1.1 for this object.

mojanm:
@mojanm will submit the necessary PR to create a new version of Single Phase Power Meter object with Tension Resource type changed to float. As Tension is a Mandatory resource, the new version of Object will be 2.0.

seanmcilroy29:
Object updated - See https://github.com/OpenMobileAlliance/lwm2m-registry/pull/582

PR #579, opened 2021-02-02
Update 10243.xml to addres issue #567

mojanm:
Changing resource 4 type to address Issue #567

new pull request template


mojanm:
Need to wait for updated the DDF file

Issue #577, opened 2021-02-01
Typo in Actuation Object


akeranen:
Resource 5853 ("Muti-state Output") of Actuation (3306) Object should be "Multi-state Output"

https://github.com/OpenMobileAlliance/lwm2m-registry/blob/prod/3306.xml#L76

jpradocueva:
Maintainer to create a Pull Request to be reviewed next week.

seanmcilroy29:
Issue resolved - see PR https://github.com/OpenMobileAlliance/lwm2m-registry/pull/584

PR #584, opened 2021-02-11
[Admin] Spelling error corrected 5853 updated

seanmcilroy29:
Spelling error corrected in RR 5853 `muti-state `to `multi-state`
Effects 1 Object - 3306

Resolves Issue https://github.com/OpenMobileAlliance/lwm2m-registry/issues/577


jpradocueva (changes requested):
There are two more versions of this object in the "version_history" folder that need to be updated too.

Issue #583, opened 2021-02-10
Duplicate Timestamp resource in Location object description (version 1.1)


rlubos:
The Timestamp resource (5518) is duplicated in the description of the Location object (3336) in version 1.1. It was already present in object version 1.0 and now there are two entries for the resource, first at [line 96](https://github.com/OpenMobileAlliance/lwm2m-registry/blob/prod/3336.xml#L96) and then at [line 116](https://github.com/OpenMobileAlliance/lwm2m-registry/blob/prod/3336.xml#L116).

Also, the indentation of the resources added in object version 1.1 seems to be one tab more compared to the other resources, not sure if that's intended?

sbernard31:
I confirm this :

Leshan raises : 
```
Object 3336 in /models/3336.xml contains at least 2 resources with same id 5518.
```

mkgillmore:
We confirm this issue and will attempt to address ASAP

PR #588, opened 2021-02-23
[Admin] Object 3336 - Duplicate resource 5518 removed

seanmcilroy29:
Object 3336 - Duplicate resource 5518 removed - See issue https://github.com/OpenMobileAlliance/lwm2m-registry/issues/583

### DIFF
--- a/3336.xml
+++ b/3336.xml
@@ -112,47 +112,37 @@ POSSIBILITY OF SUCH DAMAGE.
                 <RangeEnumeration></RangeEnumeration>
                 <Units></Units>
                 <Description>The application type of the sensor or actuator as a string depending on the use case.</Description>
-            </Item>
-			<Item ID="5518">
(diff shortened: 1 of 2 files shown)


jpradocueva (changes requested):
We need to do the same changes in 3336-v1_0 and 3336-v1_1.

seanmcilroy29:
Version History updated;
3336-v1_0 doesn't have the duplicated RR
3336-v1_1 updated

Validated with no errors

jpradocueva:
@seanmcilroy29 it makes me wonder how is possible in that case that the 3336 v1.1 had introduced a duplicate in the first place. Can we check the history and see if this is an isolated issue or a wide spread problem, may be introduced by the tool that creates objects for the "version_history"?

jpradocueva (approved):
Ok

PR #592, opened 2021-02-26
[ADMIN] Aligning test with dev updates

seanmcilroy29:
Aligning test with dev updates


jpradocueva (approved):
It is not possible to reproduce the original validation issue.
We have validated the PR in several occasions and it validates correctly. 
The original validation problem may have been created by a bad connection at the time of validation.

PR #604, opened 2021-03-29
Create 505.xml

asoloway64:
Need to separate the new object 504 which is for Remote SIM Provisioning directly to a SM-DP+ server and this new object which is Remote SIM Provisioning through a LwM2M server (downloading the appropriate package).

new pull request template

### DIFF
--- a/505.xml
+++ b/505.xml
@@ -226,7 +226,7 @@ LwM2M Client.]]></Description>
 				<Mandatory>Optional</Mandatory>
 				<Type>Integer</Type>
 				<RangeEnumeration></RangeEnumeration>
-				<Units>bit</Units>
+				<Units>KiB</Units>
 				<Description><![CDATA[Total amount of storage space which can store data and software in the LwM2M Device.]]></Description>
 			</Item>
 			<Item ID="14">
--- a/505.xml
+++ b/505.xml
@@ -1,0 +1,279 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<!-- 
+FILE INFORMATION
+OMA Permanent Document
+   File: OMA-SUP-XML_505-V1_0-YYYY-MM-DD-A
+   Path: https://www.openmobilealliance.org/release/ObjLwM2M_Remote_SIM_Provisioning
+Public Reachable Information
+   Path: https://github.com/OpenMobileAlliance/lwm2m-registry
+   Name: 505.xml
+NORMATIVE INFORMATION
+  Send comments to https://github.com/OpenMobileAlliance/OMA_LwM2M_for_Developers/issues
+LEGAL DISCLAIMER
+Copyright 2021 Open Mobile Alliance. 
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions
+are met:
+1. Redistributions of source code must retain the above copyright
+notice, this list of conditions and the following disclaimer.
+2. Redistributions in binary form must reproduce the above copyright
+notice, this list of conditions and the following disclaimer in the
+documentation and/or other materials provided with the distribution.
+3. Neither the name of the copyright holder nor the names of its
+contributors may be used to endorse or promote products derived
+from this software without specific prior written permission.
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+"AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+POSSIBILITY OF SUCH DAMAGE.
+The above license is used as a license under copyright only.  Please
+reference the OMA IPR Policy for patent licensing terms:
+https://www.omaspecworks.org/about/intellectual-property-rights/
+-->
+
+<LWM2M xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="http://openmobilealliance.org/tech/profiles/LWM2M-v1_1.xsd">
+<Object ObjectType="MODefinition">
+		<Name>Remote SIM Provisioning</Name>
+		<Description1><![CDATA[This is a device management object that should be used for Remote SIM Provisioning from a LwM2M server]]></Description1>
+		<ObjectID>505</ObjectID>
+		<ObjectURN>urn:oma:lwm2m:oma:505</ObjectURN>
+		<LWM2MVersion>1.0</LWM2MVersion>
+		<ObjectVersion>1.0</ObjectVersion>
+		<MultipleInstances>Multiple</MultipleInstances>
+		<Mandatory>Optional</Mandatory>
+		<Resources>
+			<Item ID="0">
+				<Name>Current UIM Type</Name>
+				<Operations>R</Operations>
+				<MultipleInstances>Single</MultipleInstances>
+				<Mandatory>Mandatory</Mandatory>
+				<Type>Integer</Type>
+				<RangeEnumeration></RangeEnumeration>
+				<Units></Units>
+				<Description><![CDATA[Provides the information about the currently being used SIM Type:
+0: Full Size
+1: Mini Size
+2: Micro
+3: Nano
+4: eSIM
+5: UICC
+6: Soft SIM
+7: Thin SIM
+8: eUICC
+9: iSIM
+10: iUICC
+11-24: Reserved for future use]]></Description>
+			</Item>
+			<Item ID="1">
+				<Name>Supported SIM Type</Name>
+				<Operations>R</Operations>
+				<MultipleInstances>Single</MultipleInstances>
+				<Mandatory>Mandatory</Mandatory>
+				<Type>Integer</Type>
+				<RangeEnumeration></RangeEnumeration>
+				<Units></Units>
+				<Description><![CDATA[Provides the information about the currently supported SIM Types: 
+0: Full Size
+1: Mini Size
+2: Micro
+3: Nano
+4: eSIM
+5: UICC
+6: Soft SIM
+7: Thin SIM
+8: eUICC
+9: iSIM
+10: iUICC
+11-24: Reserved for future use]]></Description>
+			</Item>
+			<Item ID="2">
+				<Name>Service Provider Name</Name>
+				<Operations>R</Operations>
+				<MultipleInstances>Single</MultipleInstances>
+				<Mandatory>Mandatory</Mandatory>
+				<Type>String</Type>
+				<RangeEnumeration></RangeEnumeration>
+				<Units></Units>
+				<Description><![CDATA[Provides the Service Provider Name using which LwM2M is currently communicating to the Server]]></Description>
+			</Item>
+			<Item ID="3">
+				<Name>Profile Package</Name>
+				<Operations>W</Operations>
+				<MultipleInstances>Single</MultipleInstances>
+				<Mandatory>Mandatory</Mandatory>
+				<Type>Opaque</Type>
+				<RangeEnumeration></RangeEnumeration>
+				<Units></Units>
+				<Description><![CDATA[Profile Package is a personalised Profile using an interoperable description format that is transmitted to an eUICC to load and install a Profile.]]></Description>
+			</Item>
+			<Item ID="4">
+				<Name>Profile URI</Name>
+				<Operations>RW</Operations>
+				<MultipleInstances>Single</MultipleInstances>
+				<Mandatory>Mandatory</Mandatory>
+				<Type>String</Type>
+				<RangeEnumeration>0..255</RangeEnumeration>
+				<Units></Units>
+				<Description><![CDATA[URI from where the device can download the profile package by an alternative mechanism. As soon the device has received the Profile package URI it performs the download at the next practical opportunity. The URI format is defined in RFC 3986. For example, coaps://example.org/profile is a syntactically valid URI. The URI scheme determines the protocol to be used. For CoAP this endpoint MAY be a LwM2M Server but does not necessarily need to be. A CoAP server implementing block-wise transfer is sufficient as a server hosting a firmware repository and the expectation is that this server merely serves as a separate file server making profile images available to LwM2M Clients. This server can be the future carrier server as well from which IoT devices would like to use the service.]]></Description>
+			</Item>
+			<Item ID="5">
+				<Name>Profile Update</Name>
+				<Operations>E</Operations>
+				<MultipleInstances>Single</MultipleInstances>
+				<Mandatory>Optional</Mandatory>
+				<Type></Type>
+				<RangeEnumeration></RangeEnumeration>
+				<Units></Units>
+				<Description><![CDATA[Updates profile by using the profile package stored in Package, or, by using the profile downloaded from the Package URI. This Resource is only executable when the value of the State Resource is Downloaded.]]></Description>
+			</Item>
+			<Item ID="6">
+				<Name>State</Name>
+				<Operations>R</Operations>
+				<MultipleInstances>Single</MultipleInstances>
+				<Mandatory>Mandatory</Mandatory>
+				<Type>Integer</Type>
+				<RangeEnumeration>0..3</RangeEnumeration>
+				<Units></Units>
+				<Description><![CDATA[Indicates current state with respect to this profile update. This value is set by the LwM2M Client. 
+0: Idle (before downloading or after successful updating) 
+1: Downloading (The data sequence is on the way) 
+2: Downloaded 
+3: Updating If writing the profile package to Package Resource is done, or, if the device has downloaded the profile package from the Package URI the state changes to Downloaded. Writing an empty string to Package URI Resource or setting the Package Resource to NULL (â€˜\0â€™), resets the Profile Update State Machine: the State Resource value is set to Idle and the Update Result Resource value is set to 0. When in Downloaded state, and the executable Resource Update is triggered, the state changes to Updating. If the Update Resource failed, the state returns at Downloaded. If performing the Update Resource was successful, the state changes from Updating to Idle.]]></Description>
+			</Item>
+			<Item ID="7">
+				<Name>Update Result</Name>
+				<Operations>R</Operations>
+				<MultipleInstances>Single</MultipleInstances>
+				<Mandatory>Mandatory</Mandatory>
+				<Type>Integer</Type>
+				<RangeEnumeration>0..9</RangeEnumeration>
+				<Units></Units>
+				<Description><![CDATA[Contains the result of downloading or updating the Profile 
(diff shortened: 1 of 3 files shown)


505.xml
```diff
@@ -0,0 +1,265 @@
...
+		<MultipleInstances>Multiple</MultipleInstances>
+		<Mandatory>Optional</Mandatory>
+		<Resources>
+			<Item ID="0">
+				<Name>Current SIM Type</Name>
+				<Operations>R</Operations>
+				<MultipleInstances>Single</MultipleInstances>
+				<Mandatory>Mandatory</Mandatory>
+				<Type>Integer</Type>
+				<RangeEnumeration></RangeEnumeration>
```
akeranen: `0..24` appropriate here? Or even `0..3` since you can update this when you make new version of the object to add new values. Same applies to all the other enumerations too.

505.xml
```diff
@@ -0,0 +1,265 @@
...
+			</Item>
+			<Item ID="2">
+				<Name>Service Provider Name</Name>
+				<Operations>R</Operations>
+				<MultipleInstances>Single</MultipleInstances>
+				<Mandatory>Mandatory</Mandatory>
+				<Type>String</Type>
+				<RangeEnumeration></RangeEnumeration>
+				<Units></Units>
+				<Description><![CDATA[Provides the Service Provider Name using which LwM2M is currently communicating to the Server]]></Description>
```
akeranen: Full stops missing from the end of many descriptions.

505.xml
```diff
@@ -0,0 +1,265 @@
...
+				<Operations>W</Operations>
+				<MultipleInstances>Single</MultipleInstances>
+				<Mandatory>Optional</Mandatory>
+				<Type>Opaque</Type>
+				<RangeEnumeration></RangeEnumeration>
+				<Units></Units>
+				<Description><![CDATA[Profile Package is a personalised Profile using an interoperable description format that is transmitted to an eUICC to load and install a Profile.]]></Description>
+			</Item>
+			<Item ID="4">
+				<Name>Profile URI</Name>
```
akeranen, line 102: Should this be URL? For example URNs are probably not OK?

505.xml
```diff
@@ -0,0 +1,265 @@
...
+				<Operations>RW</Operations>
+				<MultipleInstances>Single</MultipleInstances>
+				<Mandatory>Optional</Mandatory>
+				<Type>String</Type>
+				<RangeEnumeration>0..255</RangeEnumeration>
+				<Units></Units>
+				<Description><![CDATA[URI from where the device can download the profile package by an alternative mechanism. As soon the device has received the Profile package URI it performs the download at the next practical opportunity. The URI format is defined in RFC 3986. For example, coaps://example.org/profile is a syntactically valid URI. The URI scheme determines the protocol to be used. For CoAP this endpoint MAY be a LwM2M Server but does not necessarily need to be. A CoAP server implementing block-wise transfer is sufficient as a server hosting a firmware repository and the expectation is that this server merely serves as a separate file server making profile images available to LwM2M Clients. This server can be the future carrier server as well from which IoT devices would like to use the service.]]></Description>
+			</Item>
+			<Item ID="5">
+				<Name>Profile Update</Name>
```
akeranen: Maybe "Update Profile" is better name since it's action

505.xml
```diff
@@ -0,0 +1,265 @@
...
+2: Downloaded 
+3: Updating If writing the profile package to Package Resource is done, or, if the device has downloaded the profile package from the Package URI the state changes to Downloaded. Writing an empty string to Package URI Resource or setting the Package Resource to NULL (â€˜\0â€™), resets the Profile Update State Machine: the State Resource value is set to Idle and the Update Result Resource value is set to 0. When in Downloaded state, and the executable Resource Update is triggered, the state changes to Updating. If the Update Resource failed, the state returns at Downloaded. If performing the Update Resource was successful, the state changes from Updating to Idle.]]></Description>
+			</Item>
+			<Item ID="7">
+				<Name>Update Result</Name>
+				<Operations>R</Operations>
+				<MultipleInstances>Single</MultipleInstances>
+				<Mandatory>Mandatory</Mandatory>
+				<Type>Integer</Type>
+				<RangeEnumeration>0..9</RangeEnumeration>
```
akeranen: ..8 ?

505.xml
```diff
@@ -0,0 +1,265 @@
...
+			<Item ID="10">
+				<Name>Profile Update Protocol Support</Name>
+				<Operations>R</Operations>
+				<MultipleInstances>Multiple</MultipleInstances>
+				<Mandatory>Optional</Mandatory>
+				<Type>Integer</Type>
+				<RangeEnumeration></RangeEnumeration>
+				<Units></Units>
+				<Description><![CDATA[This resource indicates what protocols the LwM2M Client implements to retrieve Profiles. The LwM2M server uses this information to decide what URI to include in the Package URI. A LwM2M Server MUST NOT include a URI in the Package URI object that uses a protocol that is unsupported by the LwM2M client. For example, if a LwM2M client indicates that it supports CoAP and CoAPS then a LwM2M Server must not provide an HTTP URI in the Packet URI. The following values are defined by this version of the specification: 
+0 - CoAP (as defined in RFC 7252) with the additional support for block-wise transfer. CoAP is the default setting.
```
akeranen:
```suggestion
0: CoAP (as defined in RFC 7252) with the additional support for block-wise transfer. CoAP is the default setting. 
```

akeranen: ..and same change in all enum items to align with other descriptions

505.xml
```diff
@@ -0,0 +1,265 @@
...
+				<Mandatory>Mandatory</Mandatory>
+				<Type>Integer</Type>
+				<RangeEnumeration></RangeEnumeration>
+				<Units></Units>
+				<Description><![CDATA[The LwM2M Client uses this resource to
+indicate its support for transferring Profile
+images to the client either via the Package
+Resource (=push) or via the Package URI
+Resource (=pull) mechanism.
+0 - Pull only
```
akeranen:
```suggestion
0: Pull only
```

akeranen: etc.

505.xml
```diff
@@ -0,0 +1,265 @@
...
+LwM2M Client.]]></Description>
+			</Item>
+			<Item ID="12">
+				<Name>Free Memory on SIM</Name>
+				<Operations>R</Operations>
+				<MultipleInstances>Single</MultipleInstances>
+				<Mandatory>Optional</Mandatory>
+				<Type>Integer</Type>
+				<RangeEnumeration></RangeEnumeration>
+				<Units>bit</Units>
```
akeranen:
```suggestion
				<Units>KiB</Units>
```

505.xml
```diff
@@ -0,0 +1,265 @@
...
+				<Description><![CDATA[Estimated current available amount of storage space on SIM which can store data and software in the LwM2M Device (expressed in kilobytes).]]></Description>
+			</Item>
+			<Item ID="13">
+				<Name>Total Memory on SIM</Name>
+				<Operations>R</Operations>
+				<MultipleInstances>Single</MultipleInstances>
+				<Mandatory>Optional</Mandatory>
+				<Type>Integer</Type>
+				<RangeEnumeration></RangeEnumeration>
+				<Units>bit</Units>
```
akeranen: wrong unit here too?

505.xml
akeranen: KiB is the unit for kibibytes (1024 * bytes). Or if this should be kilobytes (1000 * bytes) we could register that unit too.

akeranen (changes requested):

505.xml
```diff
@@ -0,0 +1,265 @@
...
+ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+POSSIBILITY OF SUCH DAMAGE.
+The above license is used as a license under copyright only.  Please
+reference the OMA IPR Policy for patent licensing terms:
+https://www.omaspecworks.org/about/intellectual-property-rights/
+-->
+
+<LWM2M xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="http://openmobilealliance.org/tech/profiles/LWM2M-v1_1.xsd">
+<Object ObjectType="MODefinition">
+		<Name>LwM2M Remote SIM Provisioning</Name>
```
PerStahl: The object name "LwM2M Remote SIM Provisioning" does not capture the specific use of this object for the nuSIM Remote SIM Provisioning solution. The use of LwM2M server is not limited to nuSIM but is also used for profile management in GSMA eSIM as covered by object 504.xml. We need a different naming in order to make clear what is the difference between these two objects (504 and 505).

505.xml
```diff
@@ -0,0 +1,265 @@
...
+				<Name>eUICC ID</Name>
+				<Operations>R</Operations>
+				<MultipleInstances>Single</MultipleInstances>
+				<Mandatory>Optional</Mandatory>
+				<Type>String</Type>
+				<RangeEnumeration></RangeEnumeration>
+				<Units></Units>
+				<Description><![CDATA[eUICC-ID (a.k.a. EID), see GSMA SGP.02 and GSMA SGP.29 for definitions]]></Description>
+			</Item>
+			<Item ID="16">
```
PerStahl, line 249: This is the profile type in consumer variant given on ES2+ I/F when indicating to SM-DP+ what type of profile to create. How is this information is encoded in the profile package such that it can be provided from the LwM2M client to the LwM2M server? Can you please provide information on the foreseen usage of this resource?
There is a reference [4] in the description of the resource. What is it referring to?

505.xml
PerStahl: Since this object and its resources is specific to the nuSIM use case and behavior we suggest to call this object "nuSIM Provisioning" with the following description:
"This is a device management object that should be used for nuSIM Provisioning".

505.xml
PerStahl: It is not clear how the profile type information is provided to the device/eUICC such that a LwM2M server can request this information from the LwM2M client. This is not something specified in SGP.22 at least, but in TrustedConnectivityAlliance specification "eUICC Profile Package: Interoperable Format Technical Specification" there is an OPTIONAL field in ProfileHeader called "profileType" with the description "The "profileType" is a free optional text indicating for example, the name of the Profile issuer and the type of Profile.". As far as I can understand this string is not necessarily the same as the Profile Type in SGP.22. If it is the profileType field of ProfileHeader that is to be returned by this resource then it is better to refer to TCA spec rather then SGP.22, I think. Given an GSMA compliant eUICC I am not sure if/how this information is stored after profile decryption and installation and how you can obtain this information. I believe also nuSIM is reusing ProfileHeader structure from TCA spec, and maybe in nuSIM case it is possible to read out this field from nuSIM.

505.xml
```diff
@@ -0,0 +1,265 @@
...
+				<Name>Profile Type</Name>
+				<Operations>R</Operations>
+				<MultipleInstances>Single</MultipleInstances>
+				<Mandatory>Optional</Mandatory>
+				<Type>Integer</Type>
```
PerStahl: This resource has type Integer, but is this really accurate? In SGP.22 it is specified as a string and also in TrustedConnectivityAlliance specification "eUICC Profile Package: Interoperable Format Technical Specification".

505.xml
asoloway64: Since this is not just about nuSIM Provisioning, we would prefer to keep the name scope in the context of LwM2M. Since it is dealing with creation of the SIM provisioning information that then gets used locally within the LwM2M client, that is capturing the concept of "Remote Provisioning". Unless there are significant objections, we would prefer to stay with our currently proposed object name.

505.xml
asoloway64: Changed to 0..3.

505.xml
asoloway64: Updated some ranges and added full stops.

505.xml
asoloway64: Used URI to align with FUMO object.

505.xml
asoloway64: Changed to "Update Profile".

505.xml
asoloway64: Changed to KiB.

505.xml
asoloway64: Change to "String"

505.xml
asoloway64: We are just trying mirror what an SM-DP+ server would be providing. Do you think we need to call out the TCA spec to do that? Can we just say it is a "free optional text" field that is produced by the SM-DP+ server and consumed by the LwM2M client?

asoloway64: According to SGP.21, Annex B: "The Operator defines its different Profile types (identified by a [non-standardised] Profile Description ID) which contains the Network Access Application like USIM,
file structure, data and applications, etc", so this is really operator specific. As we indicated, we are just duplicating what the current SM-DP+ server usage is.

505.xml
PerStahl: The term "Remote SIM Provisioning" includes both SIM profile download and SIM profile management. Clearly, in both object 504 and 505 LwM2M is used to initiate profile download and for profile management. For this reason, by using "LwM2M" in the object name will be misleading for the user that is searching for a Remote SIM Provisioning object to use with LwM2M. Therefore, we cannot accept usage of "LwM2M" in the object name. Our proposal is "nuSIM Provisioning" because this is the only Remote SIM Provisioning solution known to us that fits the proposed set of resources.

505.xml
PerStahl: Indeed this is Operator specific information that is encrypted by the SM-DP+ and is decrypted in the eUICC/iUICC/nuSIM... Exactly what is written into this field seems to be SM-DP+ specific.
The LwM2M client may obtain this information from the eUICC/iUICC/nuSIM if an appropriate such interface exists which depends on the eUICC/iUICC/nuSIM realization (not sure that is specified in GSMA eSIM). "Free optional text produced by the SM-DP+" would be OK description.

asoloway64: I changed the first sentence of the description to "Free optional text produced by the SM-DP+"

505.xml
PerStahl: The term "Remote SIM Provisioning" includes both SIM profile download and SIM profile management. Clearly, in both object 504 and 505 LwM2M is used to initiate profile download and for profile management. For this reason, by using "LwM2M" in the object name will be misleading for the user that is searching for a Remote SIM Provisioning object to use with LwM2M. Therefore, we cannot accept usage of "LwM2M" in the object name. Our proposal is "nuSIM Provisioning" because this is the only Remote SIM Provisioning solution known to us that fits the proposed set of resources.

asoloway64: OK, we are trying a new one: "LwM2M nuSIM Provisioning".  If that works for you, please mark this conversation as resolved,'

PerStahl: We are OK with the object name "LwM2M nuSIM Provisioning". However, I am not allowed to resolve this conversation as I do not have write access.

asoloway64:
I believe we are now ready to approve this object.

505.xml
```diff
@@ -0,0 +1,265 @@
...
+				<Name>Profile Update Protocol Support</Name>
+				<Operations>R</Operations>
+				<MultipleInstances>Multiple</MultipleInstances>
+				<Mandatory>Optional</Mandatory>
+				<Type>Integer</Type>
+				<RangeEnumeration>0..3</RangeEnumeration>
+				<Units></Units>
+				<Description><![CDATA[This resource indicates what protocols the LwM2M Client implements to retrieve Profiles. The LwM2M server uses this information to decide what URI to include in the Package URI. A LwM2M Server MUST NOT include a URI in the Package URI object that uses a protocol that is unsupported by the LwM2M client. For example, if a LwM2M client indicates that it supports CoAP and CoAPS then a LwM2M Server must not provide an HTTP URI in the Packet URI. The following values are defined by this version of the specification: 
+0: CoAP (as defined in RFC 7252) with the additional support for block-wise transfer. CoAP is the default setting. 
+1 - CoAPS (as defined in RFC 7252) with the additional support for block-wise transfer
```
akeranen:
```suggestion
1: CoAPS (as defined in RFC 7252) with the additional support for block-wise transfer 
```

505.xml
```diff
@@ -0,0 +1,265 @@
...
+				<Operations>R</Operations>
+				<MultipleInstances>Multiple</MultipleInstances>
+				<Mandatory>Optional</Mandatory>
+				<Type>Integer</Type>
+				<RangeEnumeration>0..3</RangeEnumeration>
+				<Units></Units>
+				<Description><![CDATA[This resource indicates what protocols the LwM2M Client implements to retrieve Profiles. The LwM2M server uses this information to decide what URI to include in the Package URI. A LwM2M Server MUST NOT include a URI in the Package URI object that uses a protocol that is unsupported by the LwM2M client. For example, if a LwM2M client indicates that it supports CoAP and CoAPS then a LwM2M Server must not provide an HTTP URI in the Packet URI. The following values are defined by this version of the specification: 
+0: CoAP (as defined in RFC 7252) with the additional support for block-wise transfer. CoAP is the default setting. 
+1 - CoAPS (as defined in RFC 7252) with the additional support for block-wise transfer 
+2 - HTTP 1.1 (as defined in RFC 7230)
```
akeranen:
```suggestion
2: HTTP 1.1 (as defined in RFC 7230) 
```

505.xml
```diff
@@ -0,0 +1,265 @@
...
+				<MultipleInstances>Multiple</MultipleInstances>
+				<Mandatory>Optional</Mandatory>
+				<Type>Integer</Type>
+				<RangeEnumeration>0..3</RangeEnumeration>
+				<Units></Units>
+				<Description><![CDATA[This resource indicates what protocols the LwM2M Client implements to retrieve Profiles. The LwM2M server uses this information to decide what URI to include in the Package URI. A LwM2M Server MUST NOT include a URI in the Package URI object that uses a protocol that is unsupported by the LwM2M client. For example, if a LwM2M client indicates that it supports CoAP and CoAPS then a LwM2M Server must not provide an HTTP URI in the Packet URI. The following values are defined by this version of the specification: 
+0: CoAP (as defined in RFC 7252) with the additional support for block-wise transfer. CoAP is the default setting. 
+1 - CoAPS (as defined in RFC 7252) with the additional support for block-wise transfer 
+2 - HTTP 1.1 (as defined in RFC 7230) 
+3 - HTTPS 1.1 (as defined in RFC 7230) Additional values MAY be defined in the future. Any value not understood by the LwM2M Server MUST be ignored.]]></Description>
```
akeranen:
```suggestion
3: HTTPS 1.1 (as defined in RFC 7230) Additional values MAY be defined in the future. Any value not understood by the LwM2M Server MUST be ignored.]]></Description>
```

505.xml
```diff
@@ -0,0 +1,265 @@
...
+				<Type>Integer</Type>
+				<RangeEnumeration>0..2</RangeEnumeration>
+				<Units></Units>
+				<Description><![CDATA[The LwM2M Client uses this resource to
+indicate its support for transferring Profile
+images to the client either via the Package
+Resource (=push) or via the Package URI
+Resource (=pull) mechanism.
+0: Pull only
+1 - Push only
```
akeranen:
```suggestion
1: Push only
```

505.xml
```diff
@@ -0,0 +1,265 @@
...
+2 - Both. In this case the LwM2M Server
```
akeranen:
```suggestion
2: Both. In this case the LwM2M Server
```

505.xml
```diff
@@ -0,0 +1,265 @@
...
+			</Item>
+			<Item ID="14">
+				<Name>Integrated Circuit Card Identifier (ICCID)</Name>
+				<Operations>R</Operations>
+				<MultipleInstances>Multiple</MultipleInstances>
+				<Mandatory>Optional</Mandatory>
+				<Type>Integer</Type>
+				<RangeEnumeration></RangeEnumeration>
+				<Units></Units>
+				<Description><![CDATA[This EF provides a unique identification number for the UICC/Smart Cards. Please refer ETSI TS 102.22.1.]]></Description>
```
akeranen: Expand EF?

akeranen (changes requested):
Couple of nits

505.xml
```diff
@@ -0,0 +1,265 @@
...
+			</Item>
+			<Item ID="12">
+				<Name>Free Memory on SIM</Name>
+				<Operations>R</Operations>
+				<MultipleInstances>Single</MultipleInstances>
+				<Mandatory>Optional</Mandatory>
+				<Type>Integer</Type>
+				<RangeEnumeration></RangeEnumeration>
+				<Units>KiB</Units>
+				<Description><![CDATA[Estimated current available amount of storage space on SIM which can store data and software in the LwM2M Device (expressed in kilobytes).]]></Description>
```
akeranen: Do we mean here, and in the next resource, **kilo**bytes (as in 1000 * bytes) or **kibi**bytes (i.e., 1024 * bytes). I assume the latter since that's more common with storage (and what Units field currently says).

akeranen (changes requested):

PR #605, opened 2021-03-29
Update 3.xml - Band information in Device object

asoloway64:
Added supported band information for the device.

new pull request template

### DIFF
--- a/3.xml
+++ b/3.xml
@@ -4,7 +4,7 @@
 FILE INFORMATION
 
 OMA Permanent Document
-   File: OMA-SUP-XML_3-V1_2-20201110-A.xml
+   File: OMA-SUP-XML_3-V1_3-YYYYMMDD-C.xml
    Path: http://www.openmobilealliance.org/release/ObjLwM2M_Device/
 
 OMNA LwM2M Registry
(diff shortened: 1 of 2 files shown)


jpradocueva:
@asoloway64 although these changes passed validation, are applied to Object 3, Device, which is under the control of DMSE. These changes cannot be applied directly via the LwM2M-Registry.

asoloway64:
Understood, but I wasn’t sure if the DMSE and IPSO WG were going to combine efforts or how the “core” objects were going to be handled.

I created a similar PR#163 (Added band resources to device object<https://github.com/OpenMobileAlliance/objects-lwm2m/pull/163>) in the objects-lwm2m repo.

akeranen (changes requested):
How are all the lists in the new resources encoded? Single instance of Integer sounds tricky for that.

mojanm:
1. Instead of adding theses resources directly to device object, we should consider defining a new object, e.g. RF Capability or Connectivity HW Capability object, and have ExtDevInfo, resource 22, point to that object. This will allow the object to evolve over time to include other capabilities as and when required.
2. Will there be any requirement for the LwM2M server to disable/enable some of the bands where a device supports multiple technologies and bands??

asoloway64:
I have created a new Object 28 (Device RF Capabilities) in PR #608 (Added Object 28: Device RF Capabilities) for these new resources. No link is required in the Device Object since the new object is single instance and optional.

PR #606, opened 2021-03-29
Create 504.xml on new branch to avoid conflicts

asoloway64:
new pull request template

### DIFF
--- a/504.xml
+++ b/504.xml
@@ -245,8 +245,7 @@ the currently active profile is the first instance.]]></Description>
 0: No RSP support
 1: GSMA eSIM for M2M devices
 2: GSMA eSIM for consumer devices
-3: nuSIM
-4-24: Reserved for future use]]></Description>
+3-24: Reserved for future use]]></Description>
 			</Item>
 			<Item ID="14">
 				<Name>Selected ICCID for RSP operation</Name>
--- a/504.xml
+++ b/504.xml
@@ -198,7 +198,7 @@ protocol.]]></Description>
 				<Mandatory>Optional</Mandatory>
 				<Type>Integer</Type>
 				<RangeEnumeration></RangeEnumeration>
-				<Units>bit</Units>
+				<Units>KiB</Units>
 				<Description><![CDATA[Estimated current available amount of storage space on SIM which can store data and software in the LwM2M Device (expressed in kibibytes).]]></Description>
 			</Item>
 			<Item ID="10">
--- a/504.xml
+++ b/504.xml
@@ -208,7 +208,7 @@ protocol.]]></Description>
 				<Mandatory>Optional</Mandatory>
 				<Type>Integer</Type>
 				<RangeEnumeration></RangeEnumeration>
-				<Units>bit</Units>
+				<Units>KiB</Units>
 				<Description><![CDATA[Total amount of storage space which can store data and software in the LwM2M Device]]></Description>
 			</Item>
 			<Item ID="11">
--- a/504.xml
+++ b/504.xml
@@ -286,7 +286,7 @@ the currently active profile is the first instance.]]></Description>
 				<Type>String</Type>
 				<RangeEnumeration></RangeEnumeration>
 				<Units></Units>
-				<Description><![CDATA[Operator specific defined type of Profile. This is equivalent to the "Profile Description ID" as described in Annex B of GSMA SGP.21]]></Description>
+				<Description><![CDATA[Free optional text produced by the SM-DP+ server. This is equivalent to the "Profile Description ID" as described in Annex B of GSMA SGP.21]]></Description>
 			</Item>
 			<Item ID="17">
 				<Name>RSP Type</Name>
--- a/504.xml
+++ b/504.xml
@@ -2,9 +2,6 @@
 
 <!-- 
 FILE INFORMATION
-OMA Permanent Document
-   File: OMA-SUP-XML_504-V1_0-YYYY-MM-DD-A
-   Path: https://www.openmobilealliance.org/release/ObjLwM2M_Remote_SIM_Provisioning
 Public Reachable Information
    Path: https://github.com/OpenMobileAlliance/lwm2m-registry
    Name: 504.xml
--- a/504.xml
+++ b/504.xml
@@ -165,11 +165,7 @@ The URI format is defined in RFC 3986.]]></Description>
 14: Failed to disable profile.
 15: Failed to delete profile.
 16: Failed to reset memory.
-17-24: Reserved for future use
-
-A LwM2M client indicates the failure to retrieve the Profile using the URI provided in the Download URI/Information resource by writing the value 8 to the /x/0/7 (Update Result resource) when the URI contained
-a URI scheme unsupported by the client. Consequently, the LwM2M Client is unable to retrieve the Profile using the URI provided by the LwM2M Server in the Download URI/Information when it refers to an unsupported
-protocol.]]></Description>
+17-24: Reserved for future use]]></Description>
 			</Item>
 			<Item ID="7">
 				<Name>Profile Name</Name>
--- a/504.xml
+++ b/504.xml
@@ -43,7 +43,7 @@ https://www.omaspecworks.org/about/intellectual-property-rights/
 <LWM2M xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="http://openmobilealliance.org/tech/profiles/LWM2M-v1_1.xsd">
 <Object ObjectType="MODefinition">
 		<Name>GSMA Remote SIM Provisioning</Name>
-		<Description1><![CDATA[This is a device management object that should be used for Remote SIM Provisioning by an SM-DP+ server]]></Description1>
+		<Description1><![CDATA[This is a device management object that should be used for Remote SIM Provisioning according to GSMA standards]]></Description1>
 		<ObjectID>504</ObjectID>
 		<ObjectURN>urn:oma:lwm2m:oma:504</ObjectURN>
 		<LWM2MVersion>1.0</LWM2MVersion>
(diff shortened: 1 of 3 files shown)


504.xml
```diff
@@ -0,0 +1,404 @@
...
+				<Name>Profile Type</Name>
+				<Operations>R</Operations>
+				<MultipleInstances>Single</MultipleInstances>
+				<Mandatory>Optional</Mandatory>
+				<Type>Integer</Type>
+				<RangeEnumeration></RangeEnumeration>
+				<Units></Units>
+				<Description><![CDATA[Operator specific defined type of Profile. This is equivalent to the "Profile Description ID" as described in Annex B of GSMA SGP.21 [4]]]></Description>
+			</Item>
+			<Item ID="17">
```
PerStahl: This is the profile type in consumer variant given on ES2+ I/F when indicating to SM-DP+ what type of profile to create. How is this information is encoded in the profile package such that it can be provided from the LwM2M client to the LwM2M server? Can you please provide information on the foreseen usage of this resource?
There is a reference [4] in the description of the resource. What is it referring to?

504.xml
```diff
@@ -0,0 +1,404 @@
...
+ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+POSSIBILITY OF SUCH DAMAGE.
+The above license is used as a license under copyright only.  Please
+reference the OMA IPR Policy for patent licensing terms:
+https://www.omaspecworks.org/about/intellectual-property-rights/
+-->
+
+<LWM2M xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="http://openmobilealliance.org/tech/profiles/LWM2M-v1_1.xsd">
+<Object ObjectType="MODefinition">
+		<Name>SM-DP+ Remote SIM Provisioning</Name>
```
PerStahl: The object name "SM-DP+ Remote SIM Provisioning" does not capture the generic use of this object for different Remote SIM Provisioning solutions. This object covers both GSMA eSIM where an SM-DP+ server is used as provisioning server but where a LwM2M server is used for profile management. The object also covers nuSIM where LwM2M server is used both for providing a profile package and for profile management operations. In future use of GSMA eSIM for low-power IoT the profile package may likely be downloaded from SM-DP+ via a LwM2M server to the device (eUICC).

504.xml
mkgillmore: Abhi will provide feedback in a few days 4/13/2021

asoloway64: Removed the extraneous "[4]" from the description as it was erroneously copied from SGP.21.

To answer how the Profile Type is used:
According to SCG.22, “Profile Type” is equivalent to “Profile Description ID” of SGP.21 Annex B.
According to SGP.21 Annex B, “Profile Description ID” defines as follows: The Operator defines its different profile types (identified by a [non-standardised] Profile Description ID) which contains the Network Access Application like USIM, file structure, data and applications, etc.
As per SGP.22: “The Operator can provide the ICCID and/or the Profile Type. In case where the Profile Type is provided, the SM-DP+ is free to select one of the Profiles that matches the Profile Type.”

504.xml
asoloway64: We are fine with a new name. This object should address the use case where the LwM2M client connects directly to the SM-DP+ server or the use case where the LwM2M server is an intermediary between the LwM2M client and the SM-DP+ server. For the use case where the LwM2M server is the repository of the SIM Update Package and is used to download that package to the LwM2M client where the update SIM process is locally executed is addressed in Object 505.

504.xml
PerStahl: Thanks Alan! The part that MNO can present it to the SM-DP+ over ES2+ I/F instead of the ICCID is clear. What is not clear is how this information is provided to the device/eUICC such that a LwM2M server can request this information from the LwM2M client. This is not something specified in SGP.22 at least, but in TrustedConnectivityAlliance specification "eUICC Profile Package: Interoperable Format Technical Specification" there is an OPTIONAL field in ProfileHeader called "profileType" with the description "The "profileType" is a free optional text indicating for example, the name of the Profile issuer and the type of Profile.". As far as I can understand this string is not necessarily the same as the Profile Type in SGP.22. If it is the profileType field of ProfileHeader that is to be returned by this resource then it is better to refer to TCA spec rather then SGP.22, I think. Given an GSMA compliant eUICC I am not sure if/how this information is stored after profile decryption and installation and how you can obtain this information. I believe also nuSIM is reusing ProfileHeader structure from TCA spec, and maybe in nuSIM case it is possible to read out this field from nuSIM.

504.xml
PerStahl: Since this object and its resources is handling GSMA Remote SIM Provisioning use cases and behavior we suggest to call this object "GSMA Remote SIM Provisioning" with the following description:
"This is a device management object that should be used for Remote SIM Provisioning according to GSMA standards".

504.xml
asoloway64: Changed type to "String"

504.xml
asoloway64: Changed the name and description as suggested.

504.xml
asoloway64: Updated the description as suggested in object 505.

504.xml
```diff
@@ -0,0 +1,348 @@
...
+			</Item>
+			<Item ID="9">
+				<Name>Free Memory on SIM</Name>
+				<Operations>R</Operations>
+				<MultipleInstances>Single</MultipleInstances>
+				<Mandatory>Optional</Mandatory>
+				<Type>Integer</Type>
+				<RangeEnumeration></RangeEnumeration>
+				<Units>bit</Units>
+				<Description><![CDATA[Estimated current available amount of storage space on SIM which can store data and software in the LwM2M Device (expressed in kilobytes).]]></Description>
```
akeranen:
```suggestion
				<Description><![CDATA[Estimated current available amount of storage space on SIM which can store data and software in the LwM2M Device (expressed in kibibytes).]]></Description>
```

504.xml
```diff
@@ -0,0 +1,348 @@
...
+				<Description><![CDATA[Version of the downloaded profile / profile package or the profile / profile package currently being downloaded.]]></Description>
+			</Item>
+			<Item ID="9">
+				<Name>Free Memory on SIM</Name>
+				<Operations>R</Operations>
+				<MultipleInstances>Single</MultipleInstances>
+				<Mandatory>Optional</Mandatory>
+				<Type>Integer</Type>
+				<RangeEnumeration></RangeEnumeration>
+				<Units>bit</Units>
```
akeranen:
```suggestion
				<Units>KiB</Units>
```

504.xml
```diff
@@ -0,0 +1,348 @@
...
+		<MultipleInstances>Multiple</MultipleInstances>
+		<Mandatory>Optional</Mandatory>
+		<Resources>
+			<Item ID="0">
+				<Name>Current SIM Type</Name>
+				<Operations>R</Operations>
+				<MultipleInstances>Single</MultipleInstances>
+				<Mandatory>Mandatory</Mandatory>
+				<Type>Integer</Type>
+				<RangeEnumeration></RangeEnumeration>
```
akeranen:
```suggestion
				<RangeEnumeration>0..3</RangeEnumeration>
```

504.xml
```diff
@@ -0,0 +1,348 @@
...
+3: iUICC
+4-24: Reserved for future use]]></Description>
+			</Item>
+			<Item ID="1">
+				<Name>Supported SIM Type</Name>
+				<Operations>R</Operations>
+				<MultipleInstances>Multiple</MultipleInstances>
+				<Mandatory>Mandatory</Mandatory>
+				<Type>Integer</Type>
+				<RangeEnumeration></RangeEnumeration>
```
akeranen:
```suggestion
				<RangeEnumeration>0..3</RangeEnumeration>
```

504.xml
```diff
@@ -0,0 +1,348 @@
...
+10: Failed to authenticate server given by the URI.
+11: Generic download and installation error.
+12: Error in profile package format
+13: Failed to enable profile.
+14: Failed to disable profile.
+15: Failed to delete profile.
+16: Failed to reset memory.
+17-24: Reserved for future use
+
+A LwM2M client indicates the failure to retrieve the Profile using the URI provided in the Download URI/Information resource by writing the value 8 to the /x/0/7 (Update Result resource) when the URI contained
```
akeranen: > /x/0/7 (Update Result resource)

This does not seem correct?

504.xml
```diff
@@ -0,0 +1,348 @@
...
+				<Description><![CDATA[Estimated current available amount of storage space on SIM which can store data and software in the LwM2M Device (expressed in kilobytes).]]></Description>
+			</Item>
+			<Item ID="10">
+				<Name>Total Memory on SIM</Name>
+				<Operations>R</Operations>
+				<MultipleInstances>Single</MultipleInstances>
+				<Mandatory>Optional</Mandatory>
+				<Type>Integer</Type>
+				<RangeEnumeration></RangeEnumeration>
+				<Units>bit</Units>
```
akeranen:
```suggestion
				<Units>KiB</Units>
```

akeranen (changes requested):
I added these suggestions already earlier but they did not seem to be addressed here

504.xml
asoloway64: I can incorporate any change you prefer. Just let me know.

asoloway64:
I think I have addressed all of the requested changes. Please verify.

504.xml
PerStahl: I propose to remove these two sentences.

asoloway64: Done.

mkgillmore:
Group approves this pull request 5-11-2021 need to update this PR for version history

mkgillmore:
Merge agreed 5/18/2021

PR #621, opened 2021-06-15
Update DDF.xml

mojanm:
To go with PR#620

### DIFF
--- a/DDF.xml
+++ b/DDF.xml
@@ -621,7 +621,7 @@ Instances of this Object are linked from Instances of Object 0 using the OSCORE 
         <DDF>22.xml</DDF>
         <Vorto></Vorto>
         <DDFLink>1</DDFLink>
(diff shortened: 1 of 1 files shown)


jpradocueva:
@mojanm the problem is that this PR only modifies DDF.XML but the Object is still the old URN. Ideally, you should provide a single PR that modifies the DDF.xml updates Object 22.xml and adds 22-1_1.xml to the version history. The system will validate your whole content at one.
Be aware that if you modify the URL of the TS, the system will identify an error as that TS doesn't exist in the FTP.

mojanm:
@jpradocueva, Not sure how I could generate a single PR to modify DDF and add a file to the version history at the same time. Had a quick look through GitHub help, but drew blank.
There is also error 272 on the 22.xml PR itself which seems to suggest the object itself is not correct. I downloaded the object and tried to run it through validator, but saw few issues with lines that I'd not touched in my PR. Corrected them, but still  it fails?? So not sure how to proceed.

jpradocueva:
@mojanm I have sent you a meeting with a suggestion to have a short meeting. Please let me know if you would like to review the issues on a Zoom call, it will be faster

Issue #615, opened 2021-05-31
[question] Why firmware object is single instance?


SeppoTakalo:
I would like to raise a question regarding [Firmware update object 5](https://github.com/OpenMobileAlliance/lwm2m-registry/blob/prod/5.xml).

Why it is limited to single instance?

I'm working on embedded IoT device with multiple firmware images and looking for solution to update them all. If the given object id 5 would allow multiple instances, I would be able to specify for example that
5/0  would be bootloader
5/1 would be application firmware
5/2 would be modem firmware
 
Now when this is allowed, you could check each of the versions separately. For example:
```
5/0/6 = "BOOTLOADER"
5/0/7 = "v1.7.0"

5/1/6 = "application"
5/1/7 = "<git hash or version>"

5/2/6 = "modem"
5/2/7 = "1.3.0"
```

On the side note, I have been made aware that object ID 9 would allow multiple instances, but I can see that it defines uninstall and deactivation/activation as a mandatory resource. This is a requirement that not all embedded devices can implement.

So, would it be possible to modify object 5 to be a multi instance?

hannestschofenig:
The Firmware Update Object should be seen as a delivery mechanism. We agree that having multiple instances of this Object would be nice. This was overlooked when creating the Object for LwM2M 1.0. For your particular use case, the package could contain indication on the recipient of the update. Please have a look at the SUIT manifest https://datatracker.ietf.org/doc/draft-ietf-suit-manifest/.

SeppoTakalo:
Hi @hannestschofenig and thank you for your answer.

It is true that delivered package could contain metadata required to identify which target it is meant for. However, in my opinion, this feels a bit hacky for two reasons. Firstly, the firmware object exposes version string, but if you can deliver multiple images, which version this is, or should it be a combined string of all version numbers? Secondly, usually manufacturer or the update software already dictates the binary format, and might already contain verification of correct image. Feels unnecessary to pack signed image to another package which then should be processed and unpacked on limited resource device. Sometimes the components that we could update, have nothing in common. For example MCUboot could be used for application image, so it dictates which verification and signing is used for application. Then modem manufacturer have their own binary format and own signing mechanism, which it verifies.

For those two reasons, I would strongly prefer the simplicity of firmware object allowing me to transfer binaries in native format and if multiple instances would be allowed I could expose versions of each components as well.

[...]

PR #638, opened 2021-09-08
Allow multiple instances of firmware update object (5)

SeppoTakalo:
This change would allow multiple instances of firmware update object
so devices with more than one component, would allow each to be updated
separately while each of then would still expose their version information
properly.

For example:

```
5/0/6 = "BOOTLOADER"
5/0/7 = "v1.7.0"

5/1/6 = "application"
5/1/7 = "<git hash or version>"

5/2/6 = "modem"
5/2/7 = "1.3.0"
```

This change requires very minimal software changes, while enabling
usable feature.

Change is proposed in Issue #615


jpradocueva:
If the group accepts this PR, the DDF.xml needs to be updated as well as add a new object 5-1_2.xml to the "version_history" folder.

SeppoTakalo:
I have updated the PR to contain those mentioned changes to DDF.xml and version_history directory.

jpradocueva:
@SeppoTakalo the metadata also needs to be updated, e.g. the location of the file document in the FTP and the copyright. Please note that you are providing a PR directly to an Object that is internally developed by the Working Group in a private repository. The group will have to evaluate your changes and decide how to proceed. If they accept the changes as suggested, the group will have to apply these changes on their internal (private) repository.

hannestschofenig:
Hi Seppo, 

we discussed your proposal at our weekly conference call and there are two main issues: First, we have a software management object, which extends the firmware update object by adding multiple instances. In some sense this is what you are probably looking for. Second, there are concerns that this update will introduce to interoperability issues. 

For this reason, the group recommends to use the software management object (see https://raw.githubusercontent.com/OpenMobileAlliance/lwm2m-registry/prod/9.xml and http://www.openmobilealliance.org/release/LWM2M_SWMGMT/V1_0_2-20210119-A/OMA-TS-LWM2M_SwMgmt-V1_0_2-20210119-A.pdf) instead. We noticed that there is an inconsistency between the published object and what is described in the spec. We are working on an update of the spec.

SeppoTakalo:
Hi Hannes,

Thank you for your answer. As the proposal is not going to be accepted, I'll close this PR and the linked in issue.

I think we will be proceeding with Software Management object.

Few reasons why we originally tried to avoid using it are regarding the state transitions that place too strict requirements for the embedded device:
* Transition from DOWNLOADED to DELIVERED requires application to be able to verify the package. But if the package was, for example MCUBoot image, it will be verified on reboot phase. Also, if server is observing the state, this generates traffic when sending notifications of state changes.
* Uninstallation might not be possible on all scenarions. There is only one error code defined for it. I think this state transition should be optional.
* Activation - Deactivation cycle seems unnecessary for embedded devices. The resource however is mandatory.

Issue #635, opened 2021-07-27
[AVSystem] Modbus Connection, Modbus Register Cluster


kFYatek:
### 1. Choose the type of registration: ###
  - [x] ```Object``` 
  - [ ] ```Reusable Resource```
  
### 2. Choose the [Object Class](http://www.openmobilealliance.org/wp/OMNA/LwM2M/LwM2MRegistry.html#registry-table): ###
  - [ ] 3rd party Standards or Alliances
  - [x] Private companies
  - [ ] Private individuals (not allowed Reusable Resources registration)
  
## 3. Confirm Software License
  - [x] Tick this box confirming that your submission has included a [software license](http://devtoolkit.openmobilealliance.org/OEditor/License).
  

Concept            | Description
:----------------|:------------------------------
Number of ```Object(s)``` or ```Reusable Resource(s)``` to register  | 2
Name of SDO, Company or individual    | AVSystem
Object Name (as defined in the .xml file) |  Modbus Connection, Modbus Register Cluster
Short Object description, to be used in the pubic registry (max 134 characteres) (Modbus Connection) | This LwM2M Object is used to configure a Modbus protocol connection.
Short Object description, to be used in the pubic registry (max 134 characteres) (Modbus Register Cluster) | This LwM2M Object represents a set of Modbus registers and acts as an interface to read and write their values.

Proposed object definitions can be viewed at https://gist.github.com/kFYatek/2a873a83b1f8d6abb9a6415d9b78ff07. These objects are already used internally at AVSystem, with object IDs 33618 and 33619, as belonging to our bulk reservation. We would like to have them properly, publicly registered; the linked definitions tentatively use IDs 13618 and 13619 but any two (preferably consecutive) Object IDs are acceptable, since a change of pool is required anyway.

jpradocueva:
Due to lack of quorum, DMSE/IPSO postponed the review of this request for the next group meeting, Oct 19th.

jpradocueva:
@kFYatek I have created a new branch https://github.com/OpenMobileAlliance/lwm2m-registry/tree/10374_10375-AVSystem. As you can see the Objects ID for each of the registration is 10374 and 10375, the DDF.xml is updated accordingly but you will need to add the objects in the DDF property.

Please read the [README.md guidelines](https://github.com/OpenMobileAlliance/lwm2m-registry/tree/10374_10375-AVSystem#readme) to ensure a smooth validation. Ping me if you have any questions.

[...]

PR #646, opened 2021-10-20
[AVSystem] [Modbus] Add definitions of objects 10374 and 10375

kFYatek:
This contains the implementation of objects requested in issue #635.

The objects are identical to the preview previously posted in the issue, just with final object IDs and updated URNs.


mkgillmore:
Group agreed to merge 11/16/2021

PR #655, opened 2021-12-15
Nusim initial object version submissions

mkgillmore:
new pull request template

### DIFF
--- a/507.xml
+++ b/507.xml
@@ -1,0 +1,115 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- BSD-3 Clause License
+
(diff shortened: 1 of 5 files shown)


mkgillmore:
nuSIM Profile object:  This object contains all the elements for a nuSIM profile package according to nuSIM Technical Specification NST.01 and contains the actual SIM data (belonging to the MNO) in encrypted form.  It is used with the nuSIM object to enable profile changes similar to firmware updates. 

nuSIM object:  This object represents the nuSIM integrated SIM according to the nuSIM Technical Specification NST.01.  It includes the instance of the nuSIM profile object that will be loaded into the device’s nuSIM (integrated SIM).

PR #659, opened 2022-03-01
Correct to match objects-lwm2m/21.xml

jmudge:
Files seem to have been pretty-printed and don't exactly match the source file.


jpradocueva (approved):
Ok to merge

PR #660, opened 2022-03-07
Object#10378,  Edge Application Server Information

ezhaboy:
This LwM2M Object used for UE to obtain information about Edge Application Server that it shall communicate through an Edge Data Network, refer to [3GPP TS 23.558].

### DIFF
--- a/10378.xml
+++ b/10378.xml
@@ -1,0 +1,117 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!--Copyright 2022 Ericsson.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions
+are met:
+
+1. Redistributions of source code must retain the above copyright
+notice, this list of conditions and the following disclaimer.
+2. Redistributions in binary form must reproduce the above copyright
+notice, this list of conditions and the following disclaimer in the
+documentation and/or other materials provided with the distribution.
+3. Neither the name of the copyright holder nor the names of its
+contributors may be used to endorse or promote products derived
+from this software without specific prior written permission.
+
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+"AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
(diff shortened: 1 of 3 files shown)


mkgillmore (changes requested):
License is specified twice,  Please remove the redundant license declaration.

10378.xml
```diff
@@ -0,0 +1,88 @@
...
+		<ObjectURN>urn:oma:lwm2m:x:10378</ObjectURN>
+		<LWM2MVersion>1.0</LWM2MVersion>
+		<ObjectVersion>1.0</ObjectVersion>
+		<MultipleInstances>Multiple</MultipleInstances>
+		<Mandatory>Optional</Mandatory>
+		<Resources>
+			<Item ID="0">
+				<Name>EAS ID</Name>
+				<Operations>RW</Operations>
+				<MultipleInstances>Single</MultipleInstances>
```
hannestschofenig, line 47: Matt suggested to think about using multiple instances of this resource here.

asoloway64:
I think we need to have a detailed discussion on the purpose/usage of this object. 3GPP TS 23.558 specifies a complete ecosystem for enabling an Application Client (AC) in the UE to access an Edge Application Server (EAS). This object does not take into account that entire ecosystem, but uses it as a justification to provide EAS contact information. I would counter that we can propose a cloud or edge application server information object, but it should not claim to be aligned with TS 23.558.

ezhaboy:
> I think we need to have a detailed discussion on the purpose/usage of this object. 3GPP TS 23.558 specifies a complete ecosystem for enabling an Application Client (AC) in the UE to access an Edge Application Server (EAS). This object does not take into account that entire ecosystem, but uses it as a justification to provide EAS contact information. I would counter that we can propose a cloud or edge application server information object, but it should not claim to be aligned with TS 23.558.

Thank you for your comprehensive comments,  the purpose of this new Object is to provide EAS information to the LwM2M Client (as a lightweight EEC), leveraging LwM2M existing features.
In this use case, A LwM2M Server will interact with ECS and EES(s) to obtain EAS information through reference points EDGE-1 and EDGE-4 on behalf of the EEC (acting as a gateway), assisting EAS discovery and service continuity.
The Object refers terminology of [3GPP TS 23.558], it may not used to bring entire ecosystem of [3GPP TS 23.558], therefore the Object registered as company produced at current stage.

asoloway64:
I now see that this is an Ericsson specific object. I have no objections.

10378.xml
ezhaboy: @mkgillmore, @hannestschofenig as confirmed that three Resource Instance can be multiple.

ezhaboy:
@hannestschofenig, @mkgillmore  I have to on hold registration due company internal process, sorry for that.

PR #668, opened 2022-03-09
Object#10376 (Periodic Activity) 

BTE164:
Generic interface to set up periodic activity like hourly sensor reading/load activation and mechanism to retrieve the periodic data.

### DIFF
--- a/10376.xml
+++ b/10376.xml
@@ -133,7 +133,7 @@ Start Mask is adjusted to UTC offset provided by Device [/3/0/14] object instanc
 				<Type>Opaque</Type>
 				<RangeEnumeration></RangeEnumeration>
 				<Units></Units>
-				<Description><![CDATA[Record resource enable the accumulation of returned data(if any) when the activity is run/ called. The accumulated data or records is represented by a read-only serialised Opaque (Binary).
+				<Description><![CDATA[Record resource enable the accumulation of returned data (if any) when the activity is run/called. The accumulated data or records is represented by a read-only serialized Opaque (Binary) value.
 The number of records available in this serialised resource can be calculated by Record Head Index - Record Current Read Index. Nevertheless, the number of records serialised is limited by device's memory allocation. After successful record read (manual read, observation or dispatch), device will increment Record Current Read Index as per how many already read.
 
 The record use CBOR format encoded as following:
--- a/10376.xml
+++ b/10376.xml
@@ -60,7 +60,7 @@ POSSIBILITY OF SUCH DAMAGE.
 				<Type>String</Type>
 				<RangeEnumeration></RangeEnumeration>
 				<Units></Units>
-				<Description><![CDATA[This resource describes activity's operation including it's supported options and data returned.]]>
+				<Description><![CDATA[This resource describes activity's operation including its supported options and data returned.]]>
 				</Description>
 			</Item>
 			<Item ID="2">
--- a/10376.xml
+++ b/10376.xml
@@ -191,7 +191,7 @@ Record Timestamp is a 32-bit integer representing the number of seconds since Ja
 
 Dispatch Start Mask is adjusted to UTC offset provided by Device [/3/0/14] object instance resource, or adjusted to UTC if the resource is not provided.
 
-If this resource is used together with Dispatch Interval resource, it is possible to automatically deliver Record resource at specific time. I.e.  dispatch accumulated record every 6 hours, at minute 5 (00:05, 06:05, 12:05, 18:05).
+If this resource is used together with Dispatch Interval resource, it is possible to automatically deliver Record resource at specific time. For example, dispatch accumulated record every 6 hours, at minute 5 (00:05, 06:05, 12:05, 18:05).
 ]]></Description>
 			</Item>
 			<Item ID="11">
--- a/10376.xml
+++ b/10376.xml
@@ -106,7 +106,7 @@ POSSIBILITY OF SUCH DAMAGE.
 				<Mandatory>Mandatory</Mandatory>
 				<Type>Integer</Type>
 				<RangeEnumeration></RangeEnumeration>
-				<Units></Units>
+				<Units>s</Units>
 				<Description><![CDATA[This resource represents the time interval between the run of consecutive activities. The next activity will run after Periodic Interval resource value has elapsed.]]>
 				</Description>
 			</Item>
--- a/10376.xml
+++ b/10376.xml
@@ -158,7 +158,7 @@ POSSIBILITY OF SUCH DAMAGE.
 				<Type>Integer</Type>
 				<RangeEnumeration></RangeEnumeration>
 				<Units></Units>
-				<Description><![CDATA[This resource represents the index of the oldest record stored in device. This index is adjusted(incremented) by device based on it's memory space availability.]]>
+				<Description><![CDATA[This resource represents the index of the oldest record stored in device. This index is adjusted (incremented) by device based on its memory space availability.]]>
 				</Description>
 			</Item>
 			<Item ID="9">
--- a/10376.xml
+++ b/10376.xml
@@ -49,7 +49,7 @@ POSSIBILITY OF SUCH DAMAGE.
 				<Type>String</Type>
 				<RangeEnumeration></RangeEnumeration>
 				<Units></Units>
-				<Description><![CDATA[This resource represents activity's name or string reference declared by device. It cannot be modified externally. Device can support multiple activities by declaring multiple instances of Periodic Activity object.]]>
+				<Description><![CDATA[This resource represents activity's name declared by device. It cannot be modified externally. Device can support multiple activities by declaring multiple instances of Periodic Activity object.]]>
 				</Description>
 			</Item>
 			<Item ID="1">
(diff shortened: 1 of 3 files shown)


10376.xml
```diff
@@ -0,0 +1,202 @@
...
+		<Resources>
+			<Item ID="0">
+				<Name>Activity Name</Name>
+				<Operations>R</Operations>
+				<MultipleInstances>Single</MultipleInstances>
+				<Mandatory>Mandatory</Mandatory>
+				<Type>String</Type>
+				<RangeEnumeration></RangeEnumeration>
+				<Units></Units>
+				<Description><![CDATA[This resource represents activity's name or string reference declared by device. It cannot be modified externally. Device can support multiple activities by declaring multiple instances of Periodic Activity object.]]>
```
akeranen: The meaning of "or string reference declared by device" wasn't quite clear here

10376.xml
```diff
@@ -0,0 +1,202 @@
...
+			</Item>
+			<Item ID="1">
+				<Name>Activity Description</Name>
+				<Operations>R</Operations>
+				<MultipleInstances>Single</MultipleInstances>
+				<Mandatory>Mandatory</Mandatory>
+				<Type>String</Type>
+				<RangeEnumeration></RangeEnumeration>
+				<Units></Units>
+				<Description><![CDATA[This resource describes activity's operation including it's supported options and data returned.]]>
```
akeranen:
```suggestion
				<Description><![CDATA[This resource describes activity's operation including its supported options and data returned.]]>
```

10376.xml
```diff
@@ -0,0 +1,202 @@
...
+			</Item>
+			<Item ID="3">
+				<Name>Start Mask</Name>
+				<Operations>RW</Operations>
+				<MultipleInstances>Single</MultipleInstances>
+				<Mandatory>Mandatory</Mandatory>
+				<Type>String</Type>
+				<RangeEnumeration></RangeEnumeration>
+				<Units></Units>
+				<Description><![CDATA[This resource defines when to start the periodic activity. Start Mask needs to adhere to following string format:
```
akeranen, line 85: The description field has many tab characters that may not work well with all UIs. I'd recommend using only line changes. Same comment applies to many other descriptions here.

10376.xml
```diff
@@ -0,0 +1,202 @@
...
+				</Description>
+			</Item>
+			<Item ID="4">
+				<Name>Periodic Interval</Name>
+				<Operations>RW</Operations>
+				<MultipleInstances>Single</MultipleInstances>
+				<Mandatory>Mandatory</Mandatory>
+				<Type>Integer</Type>
+				<RangeEnumeration></RangeEnumeration>
+				<Units></Units>
```
akeranen:
```suggestion
				<Units>s</Units>
```

I suppose the value is in seconds?

10376.xml
```diff
@@ -0,0 +1,202 @@
...
+				</Description>
+			</Item>
+			<Item ID="5">
+				<Name>Run Period</Name>
+				<Operations>RW</Operations>
+				<MultipleInstances>Single</MultipleInstances>
+				<Mandatory>Optional</Mandatory>
+				<Type>Integer</Type>
+				<RangeEnumeration></RangeEnumeration>
+				<Units></Units>
```
akeranen:
```suggestion
				<Units>s</Units>
```

10376.xml
```diff
@@ -0,0 +1,202 @@
...
+				</Description>
+			</Item>
+			<Item ID="11">
+				<Name>Record Dispatch Interval</Name>
+				<Operations>RW</Operations>
+				<MultipleInstances>Single</MultipleInstances>
+				<Mandatory>Optional</Mandatory>
+				<Type>Integer</Type>
+				<RangeEnumeration></RangeEnumeration>
+				<Units></Units>
```
akeranen:
```suggestion
				<Units>s</Units>
```

10376.xml
```diff
@@ -0,0 +1,202 @@
...
+			</Item>	
+			<Item ID="10">
+				<Name>Record Dispatch Start Mask</Name>
+				<Operations>RW</Operations>
+				<MultipleInstances>Single</MultipleInstances>
+				<Mandatory>Optional</Mandatory>
+				<Type>String</Type>
+				<RangeEnumeration></RangeEnumeration>
+				<Units></Units>
+				<Description><![CDATA[This resource represents the start time to deliver Record resource. If Record resource is currently being observed, this resource enables accumulation of the Record changes before being dispatch (rather than delivering Record immediatly as it changes). As per Start Mask, Dispatch Start Mask must adhere to following string format:
```
akeranen: What's the expected time zone for the time? Same applies to other time related fields. Did you consider using RFC3339?

10376.xml
```diff
@@ -0,0 +1,202 @@
...
+			</Item>
+			<Item ID="8">
+				<Name>Record Tail Index</Name>
+				<Operations>R</Operations>
+				<MultipleInstances>Single</MultipleInstances>
+				<Mandatory>Optional</Mandatory>
+				<Type>Integer</Type>
+				<RangeEnumeration></RangeEnumeration>
+				<Units></Units>
+				<Description><![CDATA[This resource represents the index of the oldest record stored in device. This index is adjusted(incremented) by device based on it's memory space availability.]]>
```
akeranen:
```suggestion
				<Description><![CDATA[This resource represents the index of the oldest record stored in device. This index is adjusted (incremented) by device based on its memory space availability.]]>
```

10376.xml
BTE164: I assumed it would be enough to refer to mandatory Device [/3/0/14] object for UTC offset (perhaps I need to mention this in the description). 
Or do you think it's better to add another optional Read resource containing device UTC offset specially for this particular instance object?

10376.xml
BTE164: I edit the xml using notepad++. I have removed all the tabs.

10376.xml
BTE164: Removed "or string reference".

10376.xml
akeranen: I think both are OK options as long as you make it explicit. Alternative would be to use RFC3339 format that supports this. Perhaps the device object UTC offset reference makes most sense.

10376.xml
```diff
@@ -0,0 +1,211 @@
...
+			</Item>
+			<Item ID="6">
+				<Name>Record</Name>
+				<Operations>R</Operations>
+				<MultipleInstances>Single</MultipleInstances>
+				<Mandatory>Optional</Mandatory>
+				<Type>Opaque</Type>
+				<RangeEnumeration></RangeEnumeration>
+				<Units></Units>
+				<Description><![CDATA[Record resource enable the accumulation of returned data(if any) when the activity is run/ called. The accumulated data or records is represented by a read-only serialised Opaque (Binary).
```
akeranen:
```suggestion
				<Description><![CDATA[Record resource enable the accumulation of returned data (if any) when the activity is run/called. The accumulated data or records is represented by a read-only serialized Opaque (Binary) value.
```

10376.xml
```diff
@@ -0,0 +1,211 @@
...
+			</Item>
+			<Item ID="9">
+				<Name>Record Read Index</Name>
+				<Operations>RW</Operations>
+				<MultipleInstances>Single</MultipleInstances>
+				<Mandatory>Optional</Mandatory>
+				<Type>Integer</Type>
+				<RangeEnumeration></RangeEnumeration>
+				<Units></Units>
+				<Description><![CDATA[This resource represents the index of the last read record. Device will increment this index upon successful record delivery to server(via manual read or scheduled dispatch). Server may modify this index to traverse within available records in device.]]>
```
akeranen:
```suggestion
				<Description><![CDATA[This resource represents the index of the last read record. Device will increment this index upon successful record delivery to server (via manual read or scheduled dispatch). Server may modify this index to traverse within available records in device.]]>
```

10376.xml
```diff
@@ -0,0 +1,211 @@
...
+				<Mandatory>Optional</Mandatory>
+				<Type>String</Type>
+				<RangeEnumeration></RangeEnumeration>
+				<Units></Units>
+				<Description><![CDATA[This resource represents the start time to deliver Record resource. If Record resource is currently being observed, this resource enables accumulation of the Record changes before being dispatch (rather than delivering Record immediately as it changes). As per Start Mask, Dispatch Start Mask must adhere to following string format:
+"YYMMDDHHmmss".
+
+Dispatch Start Mask is adjusted to UTC offset provided by Device [/3/0/14] object instance resource, or adjusted to UTC if the resource is not provided.
+
+If this resource is used together with Dispatch Interval resource, it is possible to automatically deliver Record resource at specific time. I.e.  dispatch accumulated record every 6 hours, at minute 5 (00:05, 06:05, 12:05, 18:05).
```
akeranen:
```suggestion
If this resource is used together with Dispatch Interval resource, it is possible to automatically deliver Record resource at specific time. For example, dispatch accumulated record every 6 hours, at minute 5 (00:05, 06:05, 12:05, 18:05).
```

The last part of description looks like an example instead of clarification

akeranen (commented):
Added a few more editorial suggestions but otherwise looks OK to me.

10376.xml
BTE164: I have updated the description to include a reference to the Device object's UTC Offset resource. Thanks for your input.

PR #669, opened 2022-03-09
Object#10377 (Data Monitoring) 

BTE164:
Generic interface to set up data monitoring like sensor/log data and mechanism to receive alerts.

### DIFF
--- a/10377.xml
+++ b/10377.xml
@@ -1,0 +1,247 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!--Copyright 2021 George Kent (Malaysia) Berhad. 
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions
+are met:
+
+1. Redistributions of source code must retain the above copyright
+notice, this list of conditions and the following disclaimer.
+2. Redistributions in binary form must reproduce the above copyright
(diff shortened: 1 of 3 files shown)


10377.xml
```diff
@@ -0,0 +1,247 @@
...
+			<Item ID="10">
+				<Name>Comparison B </Name>
+				<Operations>RW</Operations>
+				<MultipleInstances>Single</MultipleInstances>
+				<Mandatory>Optional</Mandatory>
+				<Type>Integer</Type>
+				<RangeEnumeration></RangeEnumeration>
+				<Units></Units>
+				<Description><![CDATA[This resource represents the logical statement to compare Data resource against Reference B resource. The statements are encoded as following:
+				0 = No Comparison
```
akeranen: I would recommend to format all these fields as `<value>: <description` to follow the style of IPSO objects. That makes it possible to parse these programmatically.

```suggestion
0: No Comparison
```

akeranen (commented):
See also formatting comments in the other PR

10377.xml
BTE164: changed to ':'

akeranen:
Looks good to me, thanks!

PR #670, opened 2022-04-12
New objects Measurement Metadata and Vendor Specific Measurement Quality Reason

dirtyha:
This is a demo pr that contains new objects described by "option-3" in Vaisala presentation in IPSO-DMSE meeting 2022-04-12.


jpradocueva:
@dirtyha Object 10379.xml failed validation. The file is not well-formed and valid.

10379.xml
```diff
@@ -0,0 +1,113 @@
...
+-->
+
+<LWM2M xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="http://openmobilealliance.org/tech/profiles/LWM2M.xsd">
+    <Object ObjectType="MODefinition">
+        <Name>Measurement Metadata</Name>
+        <Description1>This object contains measurement metadata for sensor objects. 
+The corresponding sensor is identified with an object link.
+Measurement Quality Indicator and Measurement Quality Level values provided in this object have a precedence over 
+values provided in IPSO sensor object. If multiple object instances link to the same sensor object instance then 
+the metadata provided by the largest instance id takes precedence.</Description>
```
akeranen:
```suggestion
the metadata provided by the largest instance id takes precedence.
```

10379.xml
```diff
@@ -0,0 +1,113 @@
...
+    <Resources>
+        <Item ID="0">
+            <Name>Linked Sensor</Name>
+            <Operations>R</Operations>
+            <MultipleInstances>Single</MultipleInstances>
+            <Mandatory>Mandatory</Mandatory>
+            <Type>Objlnk</Type>
+            <RangeEnumeration></RangeEnumeration>
+            <Units></Units>
+            <Description>Link to the sensor object instance that this metadata relates to.
```
akeranen:
```suggestion
            <Description>Link to the sensor object instance that this metadata relates to.</Description>
```

10379.xml
```diff
@@ -0,0 +1,113 @@
...
+                <Name>Measurement Quality Reason Code</Name>
+                <Operations>R</Operations>
+                <MultipleInstances>Multiple</MultipleInstances>
+                <Mandatory>Optional</Mandatory>
+                <Type>Integer</Type>
+                <RangeEnumeration>0..1024</RangeEnumeration>
+                <Units></Units>
+                <Description>Measurement quality degradation reason reported by a smart sensor. 
+A degraded measurement has Measurement Quality Indicator value 1-3 or Measurement Quality Level < 100.
+0: UNKNOWN Reason is unkown, 1: UNDER_RANGE Measured value is under measurement range,
```
akeranen:
```suggestion
0: UNKNOWN Reason is unknown
1: UNDER_RANGE Measured value is under measurement range 
```

I'd recommend having each value in a new line (and without commas; perhaps full stop at the end like in the other resource)

10379.xml
```diff
@@ -0,0 +1,113 @@
...
+                    <Name>Measurement Quality Indicator</Name>
+                    <Operations>R</Operations>
+                    <MultipleInstances>Single</MultipleInstances>
+                    <Mandatory>Optional</Mandatory>
+                    <Type>Integer</Type>
+                    <RangeEnumeration>0..23</RangeEnumeration>
+                    <Units></Units>
+                    <Description>Measurement quality indicator reported by a smart sensor. 
+0: UNCHECKED No quality checks were done because they do not exist or can not be applied. 
+1: REJECTED WITH CERTAINTY The measured value is invalid.
```
akeranen:
```suggestion
1: REJECTED_WITH_CERTAINTY The measured value is invalid. 
```

For consistency with resource 1, using underscores in the names makes sense (also for other values here)

10379.xml
```diff
@@ -0,0 +1,113 @@
...
+                    <Type>Integer</Type>
+                    <RangeEnumeration>0..23</RangeEnumeration>
+                    <Units></Units>
+                    <Description>Measurement quality indicator reported by a smart sensor. 
+0: UNCHECKED No quality checks were done because they do not exist or can not be applied. 
+1: REJECTED WITH CERTAINTY The measured value is invalid. 
+2: REJECTED WITH PROBABILITY The measured value is likely invalid. 
+3: ACCEPTED BUT SUSPICIOUS The measured value is likely OK. 
+4: ACCEPTED The measured value is OK. 
+5-15: Reserved for future extensions. 16-23: Vendor specific measurement quality.</Description>
```
akeranen:
```suggestion
5-15: Reserved for future extensions. 
16-23: Vendor specific measurement quality.</Description>
```

Splitting into two lines. For consistency, these could also have names similar to earlier. For example "RESERVED" and "VENDOR_SPECIFIC".

10379.xml
```diff
@@ -0,0 +1,113 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<!-- BSD-3 Clause License
+
+Copyright 2019 Vaisala.
```
akeranen:
```suggestion
Copyright 2022 Vaisala. 
```

10379.xml
```diff
@@ -0,0 +1,113 @@
...
+POSSIBILITY OF SUCH DAMAGE.
+-->
+
+<LWM2M xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="http://openmobilealliance.org/tech/profiles/LWM2M.xsd">
+    <Object ObjectType="MODefinition">
+        <Name>Measurement Metadata</Name>
+        <Description1>This object contains measurement metadata for sensor objects. 
+The corresponding sensor is identified with an object link.
+Measurement Quality Indicator and Measurement Quality Level values provided in this object have a precedence over 
+values provided in IPSO sensor object. If multiple object instances link to the same sensor object instance then
```
akeranen:
```suggestion
values provided in a sensor object. If multiple object instances link to the same sensor object instance then 
```
I think this can be applied even to other than IPSO objects

10380.xml
```diff
@@ -0,0 +1,84 @@
...
+                <Operations>R</Operations>
+                <MultipleInstances>Single</MultipleInstances>
+                <Mandatory>Mandatory</Mandatory>
+                <Type>Integer</Type>
+                <RangeEnumeration>65..1024</RangeEnumeration>
+                <Units></Units>
+                <Description>Measurement quality degradation reason reported by a smart sensor. 
+Reason codes 0-64 are generic reason codes and they are 
+defined in Measurement Metadata object XML description. 
+65-128: vendor specific block #1, 129-255: reserved block, 256-1024: vendor specific block #2</Description>
```
akeranen: Splitting to separate lines (and ending with a full stop) probably makes sense also here

akeranen (commented):
Some editorial suggestions

10379.xml
akeranen: The closing tag is provided below

10379.xml
```diff
@@ -0,0 +1,129 @@
...
+ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+POSSIBILITY OF SUCH DAMAGE.
+-->
+
+<LWM2M xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="http://openmobilealliance.org/tech/profiles/LWM2M.xsd">
+    <Object ObjectType="MODefinition">
+        <Name>Measurement Metadata</Name>
+        <Description1>This object contains measurement metadata for sensor objects. 
+The corresponding sensor is identified with an object link.
+Measurement Quality Indicator and Measurement Quality Level values provided in this object have a precedence over
```
akeranen: Tiny editorial issue: even if it makes it a bit painful to review Descriptions in Github, I think the descriptions should be in one line (i.e., no line changes) unless the description has multiple paragraphs. So I would suggest to remove the hard line changes here and in other descriptions where not needed. However, for the enum descriptions (like resource 1) they are very useful to keep one line per enum value, so please keep them there.

10379.xml
```diff
@@ -0,0 +1,129 @@
...
+                <Name>Measurement Quality Indicator</Name>
+                <Operations>R</Operations>
+                <MultipleInstances>Single</MultipleInstances>
+                <Mandatory>Optional</Mandatory>
+                <Type>Integer</Type>
+                <RangeEnumeration>0..23</RangeEnumeration>
+                <Units></Units>
+                <Description>Measurement quality indicator reported by a smart sensor.
+0: UNCHECKED No quality checks were done because they do not exist or can not be applied.
+1: REJECTED WITH CERTAINTY The measured value is invalid.
```
akeranen:
```suggestion
1: REJECTED_WITH_CERTAINTY The measured value is invalid.
```
I suggest to align all the descriptions with the style in resource 1. This style also makes it easier to automatically convert the enums with a rule "next word after the value is the `short name`".

10380.xml
```diff
@@ -0,0 +1,90 @@
...
+LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+POSSIBILITY OF SUCH DAMAGE.
+-->
+
+<LWM2M xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="http://openmobilealliance.org/tech/profiles/LWM2M.xsd">
+    <Object ObjectType="MODefinition">
+        <Name>Vendor Specific Measurement Quality Reason</Name>
+        <Description1>This object contains description for vendor specific measurement quality reason codes
+provided in object Measurement Metadata (10379). Instances of this object define the mapping of vendor
```
akeranen:
```suggestion
provided in a Measurement Metadata (10379) object. Instances of this object define the mapping of vendor 
```
This word order feels more natural to me

10379.xml
```diff
@@ -0,0 +1,129 @@
...
+        <ObjectID>10379</ObjectID>
+        <ObjectURN>urn:oma:lwm2m:x:10379</ObjectURN>
+        <LWM2MVersion>1.1</LWM2MVersion>
+        <ObjectVersion>1.0</ObjectVersion>
+        <MultipleInstances>Multiple</MultipleInstances>
+        <Mandatory>Optional</Mandatory>
+        <Resources>
+            <Item ID="0">
+                <Name>Linked Sensor</Name>
+                <Operations>R</Operations>
```
akeranen: Should the Management Server be able to change these links in some cases? If they are not part of firmware, that seems natural. In that case this could be also writable (those that don't support changing can just reject the request).

10380.xml
```diff
@@ -0,0 +1,90 @@
...
+                <Name>Vendor Specific Quality Reason Code</Name>
+                <Operations>R</Operations>
+                <MultipleInstances>Single</MultipleInstances>
+                <Mandatory>Mandatory</Mandatory>
+                <Type>Integer</Type>
+                <RangeEnumeration>65..1024</RangeEnumeration>
+                <Units></Units>
+                <Description>Measurement quality degradation reason reported by a smart sensor.
+Reason codes 0-64 are generic reason codes and they are defined in Measurement Metadata object
+XML description.
```
akeranen:
```suggestion
description.
```
Given we may some day use JSON, better to not refer to the schema format :-) Maybe it's also useful to refer to the  specific resource?

10379.xml
```diff
@@ -0,0 +1,129 @@
...
+            </Item>
+            <Item ID="1">
+                <Name>Measurement Quality Reason Code</Name>
+                <Operations>R</Operations>
+                <MultipleInstances>Multiple</MultipleInstances>
+                <Mandatory>Optional</Mandatory>
+                <Type>Integer</Type>
+                <RangeEnumeration>0..1024</RangeEnumeration>
+                <Units></Units>
+                <Description>Measurement quality degradation reason reported by a smart sensor.
```
akeranen: I wonder can we say something on what "too {much, slowly, quickly}" actually means in practice? Is it based, e.g., on limits of the physics or the sensors' earlier measurements of the same quantity? Or is it sensor/application specific?

10379.xml
```diff
@@ -0,0 +1,129 @@
...
+                <RangeEnumeration>0..1024</RangeEnumeration>
+                <Units></Units>
+                <Description>Measurement quality degradation reason reported by a smart sensor. 
+A degraded measurement has Measurement Quality Indicator value 1-3 or Measurement Quality Level less than 100.
+0: UNKNOWN Reason is unknown.
+1: UNDER_RANGE Measured value is under measurement range. 
+2: OVER_RANGE Measured value is over measurement range.
+3: NOISY Measured value seems to change too quickly. 
+4: STUCK_VALUE Measured value seems to change too slowly.
+5: STEP_TOO_HIGH Measured value has changed too much.
```
akeranen: Changed too much from the previous measurement? Given it's "step". Just to make clear the difference to "too quickly".

10379.xml
```diff
@@ -0,0 +1,129 @@
...
+                <Type>Integer</Type>
+                <RangeEnumeration>0..23</RangeEnumeration>
+                <Units></Units>
+                <Description>Measurement quality indicator reported by a smart sensor.
+0: UNCHECKED No quality checks were done because they do not exist or can not be applied.
+1: REJECTED WITH CERTAINTY The measured value is invalid.
+2: REJECTED WITH PROBABILITY The measured value is likely invalid.
+3: ACCEPTED BUT SUSPICIOUS The measured value is likely OK.
+4: ACCEPTED The measured value is OK.
+5-15: Reserved for future extensions.
```
akeranen: Like you did in the VSMQR object, perhaps it's a good policy to give short names also to the ranges. For example `RESERVED` and `VENDOR_SPECIFIC`.

10380.xml
```diff
@@ -0,0 +1,90 @@
...
+65-128: VENDOR_SPECIFIC_BLOCK_1 Vendor specific block #1.
+129-255: RESERVED_VENDOR_BLOCK_2 Reserved block 2.
```
akeranen: The purpose of this block would be useful to describe in more detail

10380.xml
```diff
@@ -0,0 +1,90 @@
...
+256-1024: VENDOR_SPECIFIC_BLOCK_2 Vendor specific block #2.
```
akeranen: Maybe also good to hint why there are two separate blocks ("with 2-byte IDs" or similar).

10380.xml
```diff
@@ -0,0 +1,90 @@
...
+                </Description>
+            </Item>
+            <Item ID="1">
+                <Name>Quality Reason Name</Name>
+                <Operations>R</Operations>
+                <MultipleInstances>Single</MultipleInstances>
+                <Mandatory>Mandatory</Mandatory>
+                <Type>String</Type>
+                <RangeEnumeration></RangeEnumeration>
```
akeranen: We could consider adding min and max limits for this since it's used as identifier. For example if this needs to be at last 1 but max 100 characters, we could use here `1..100`.

10380.xml
```diff
@@ -0,0 +1,90 @@
...
+            </Item>
+            <Item ID="2">
+                <Name>Quality Reason Description</Name>
+                <Operations>R</Operations>
+                <MultipleInstances>Single</MultipleInstances>
+                <Mandatory>Mandatory</Mandatory>
+                <Type>String</Type>
+                <RangeEnumeration></RangeEnumeration>
+                <Units></Units>
+                <Description>Description of the quality reason. This text can be used if the localization is not found.
```
akeranen:
```suggestion
                <Description>Description of the quality reason. This text can be used if a localized version is not found.
```

Maybe more clear like this?

akeranen (approved):
Set of small, mostly editorial, suggestions. Otherwise looks good to me.

10379.xml
dirtyha: That would break the validation unless we also change re-usable resource 6042 description. Also the description in all IPSO sensor objects having 6042 should be updated. If that is necessary, maybe we can do it in a new pr.

10379.xml
dirtyha: Hard line changes removed, except enum descriptions.

version_history/10379-1_0.xml
```diff
@@ -0,0 +1,129 @@
...
+                <Name>Measurement Quality Indicator</Name>
+                <Operations>R</Operations>
+                <MultipleInstances>Single</MultipleInstances>
+                <Mandatory>Optional</Mandatory>
+                <Type>Integer</Type>
+                <RangeEnumeration>0..23</RangeEnumeration>
+                <Units></Units>
+                <Description>Measurement quality indicator reported by a smart sensor.
+0: UNCHECKED No quality checks were done because they do not exist or can not be applied.
+1: REJECTED WITH CERTAINTY The measured value is invalid.
```
kFYatek: Editorial nitpick: The `10379.xml` version in the root directory has underscores here. There are also a couple of minor discrepancies in the files for 10380.

10379.xml
```diff
@@ -0,0 +1,129 @@
...
+-->
+
+<LWM2M xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="http://openmobilealliance.org/tech/profiles/LWM2M.xsd">
+    <Object ObjectType="MODefinition">
+        <Name>Measurement Metadata</Name>
+        <Description1>This object contains measurement metadata for sensor objects. 
+The corresponding sensor is identified with an object link.
+Measurement Quality Indicator and Measurement Quality Level values provided in this object have a precedence over 
+values provided in a sensor object. If multiple object instances link to the same sensor object instance then 
+the metadata provided by the largest instance id takes precedence.
```
kFYatek: As mentioned during the meeting, I think the precedence rules make things confusing for prospective implementors, and seem to encourage dubious practices. I'd suggest changing the wording to something like:

> If the LwM2M Client provides measurement quality information both inside the sensor objects and in the Measurement Metadata object, it MUST ensure that identical data is reported in corresponding object instances. The LwM2M Server SHOULD use the data from the Measurement Metadata object if both are available.
> The LwM2M Client also MUST ensure that no two Measurement Metadata object instances link to the same sensor object instance. If the LwM2M Server receives such non-conforming data, the Measurement Metadata object instance with largest Instance ID SHOULD be used.

10379.xml
kFYatek: Editorial nitpick: Measurement Quality Reason Code constants, as well Vendor Specific Quality Reason Code constants in object 10380, are named with underscores, while here there are spaces. Making things consistent either way would be nice.

10379.xml
dirtyha: I can't figure out a use case where the server would re-link measurement metadata. The metadata is a result of local processing of a measurement in the device. I can't figure why the server would re-link.

10380.xml
dirtyha: Description added.

10380.xml
dirtyha: I simplified  the block structure so that there is only one vendor specific block and a reserved block after that.

10380.xml
dirtyha: I added range 1..128

version_history/10379-1_0.xml
dirtyha: Agreed, removed underscores since they do not exist in the original re-usable resource 6042 description.

10379.xml
dirtyha: Thanks, that text sounds better than the old one. Will replace.

10379.xml
dirtyha: I agree that it would be more consistent, but we would need to make changes in many objects using 6042, as I noted above. A new dedicated pr is my suggestion, if we decide to set this right.

10379.xml
akeranen: I'm thinking of a case where new sensors are added to the system and new sensor objects are created for them by MS to enable the capability. Then, also MS setting the links might make sense.

dirtyha: Device FW needs to support adaptation to these new sensors and quality processing, if not then a FW update is needed first. Being so, I think that the FW can also create needed LwM2M objects.

kFYatek (approved):
LGTM

10379.xml
akeranen: Makes sense. Let's keep Readable-only. Avoids also error cases when can't change this.

10379.xml
akeranen: Good point; I didn't realize this change would have such a big impact. Let's park this change suggestion for the next round of updates that go across all / many IPSO objects.

10379.xml
```diff
@@ -0,0 +1,119 @@
...
+ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+POSSIBILITY OF SUCH DAMAGE.
+-->
+
+<LWM2M xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="http://openmobilealliance.org/tech/profiles/LWM2M.xsd">
+    <Object ObjectType="MODefinition">
+        <Name>Measurement Metadata</Name>
+        <Description1>This object contains measurement metadata for sensor objects. The corresponding sensor is identified with an object link. If the LwM2M Client provides measurement quality information both inside the sensor objects and in the Measurement Metadata object, it MUST ensure that identical data is reported in corresponding object instances. The LwM2M Server SHOULD use the data from the Measurement Metadata object if both are available. The LwM2M Client also MUST ensure that no two Measurement Metadata object instances link to the same sensor object instance. If the LwM2M Server receives such non-conforming data, the Measurement Metadata object instance with largest Instance ID SHOULD be used.
+        </Description1>
+        <ObjectID>10379</ObjectID>
```
akeranen: Late comment, but just realized wouldn't it make sense to have these objects in the IPSO range instead of vendor range?

dirtyha: I'm fine if we decide to move them into ext-label range together with IPSO objects, but perhaps that requires more feedback from the workgroup.

dirtyha: We discussed this in the DMSE&IPSO weekly and decided to relocate the objects into IPSO id space, respectively ids 509 and 510. I have now done this relocation and it passed validation, @jpradocueva can you check that I managed to get the license statement and DDF changes right?

mkgillmore:
group agrees to merge to prod 5/10/22

PR #672, opened 2022-05-06
Fix RangeEnumeration to specified convention

akeranen:
Two types of changes are included:
- Converted exponent notation to ".." notation
- Split units and range values

Change made in relation to issue: https://github.com/OpenMobileAlliance/LwM2M/issues/764.
Main goal is to make parsing of the RangeEnumeration field easier.

However more issues exist, which will be fixed via a new issue / PR.

### DIFF
--- a/10245.xml
+++ b/10245.xml
@@ -47,7 +47,7 @@ POSSIBILITY OF SUCH DAMAGE.
                 <MultipleInstances>Single</MultipleInstances>
                 <Mandatory>Mandatory</Mandatory>
                 <Type>Boolean</Type>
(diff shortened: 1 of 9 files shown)


akeranen:
If these changes are accepted, need to also update the versions of the objects accordingly.

akeranen:
To be replaced with separate PRs for the different set of objects.

jpradocueva:
The following companies need to be notified that their objects are being updated:

Vodafone:

- 10245, Relay Management
- 10246, CrowBox Measurements
- 10247, Neighbour Cell Report
- 10249, Connected UE Report

CloudMinds

- 10334, Robot Alarm
- 10335, Event

PR #688, opened 2022-10-24
First version of Compact Composite Object

akeranen:
DDF and other supporting files to be updated

### DIFF
--- a/511.xml
+++ b/511.xml
@@ -42,7 +42,7 @@ reference the OMA IPR Policy for patent licensing terms:
 https://www.omaspecworks.org/about/intellectual-property-rights/
 
 -->
-<LWM2M xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="http://openmobilealliance.org/tech/profiles/LWM2M.xsd">
+<LWM2M xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="http://openmobilealliance.org/tech/profiles/LWM2M-v1_1.xsd">
     <Object ObjectType="MODefinition">
         <Name>Compact Composite</Name>
         <Description1>This Object enables collecting a set of Resource values from other Objects into a single resource representation with an array of values.</Description1>
--- a/511.xml
+++ b/511.xml
@@ -1,0 +1,88 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+FILE INFORMATION
+
+OMNA LwM2M Registry
+  Path: https://github.com/OpenMobileAlliance/lwm2m-registry
+  Name: 511.xml
+
+NORMATIVE INFORMATION
+
+  Send comments to https://github.com/OpenMobileAlliance/OMA_LwM2M_for_Developers/issues
+
+LEGAL DISCLAIMER
+
+Copyright 2022 Open Mobile Alliance
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions
+are met:
+1. Redistributions of source code must retain the above copyright
+notice, this list of conditions and the following disclaimer.
+2. Redistributions in binary form must reproduce the above copyright
+notice, this list of conditions and the following disclaimer in the
+documentation and/or other materials provided with the distribution.
+3. Neither the name of the copyright holder nor the names of its
+contributors may be used to endorse or promote products derived
+from this software without specific prior written permission.
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+"AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+POSSIBILITY OF SUCH DAMAGE.
+The above license is used as a license under copyright only.  Please
+reference the OMA IPR Policy for patent licensing terms:
+https://www.omaspecworks.org/about/intellectual-property-rights/
+
+-->
+<LWM2M xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="http://openmobilealliance.org/tech/profiles/LWM2M.xsd">
+    <Object ObjectType="MODefinition">
+        <Name>Compact Composite</Name>
+        <Description1>This Object enables collecting a set of Resource values from other Objects into a single resource representation with an array of values.</Description1>
+        <ObjectID>511</ObjectID>
+        <ObjectURN>urn:oma:lwm2m:ext:511</ObjectURN>
+        <LWM2MVersion>1.1</LWM2MVersion>
+        <ObjectVersion>1.0</ObjectVersion>
+        <MultipleInstances>Multiple</MultipleInstances>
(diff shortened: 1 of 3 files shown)


511.xml
```diff
@@ -0,0 +1,89 @@
...
+            </Item>
+            <Item ID="1">
+                <Name>Compact Data</Name>
+                <Operations>RW</Operations>
+                <MultipleInstances>Single</MultipleInstances>
+                <Mandatory>Mandatory</Mandatory>
+                <Type>Opaque</Type>
+                <RangeEnumeration/>
+                <Units/>
+                <Description>CBOR array of resource values in the same order as the links in the Target Links resource or empty value if there is error (e.g., link pointing to a resource that doesn't exist or is not allowed to be read because of access control rules).</Description>
```
dnav: Instead of "empty value", you should mandate the usage of the CBOR simple [undefined](https://www.rfc-editor.org/rfc/rfc8949.html#name-floating-point-numbers-and-) (0xF7), as an empty value may be valid for Opaque or String resources.

511.xml
```diff
@@ -0,0 +1,89 @@
...
+        <Resources>
+            <Item ID="0">
+                <Name>Target Links</Name>
+                <Operations>RW</Operations>
+                <MultipleInstances>Single</MultipleInstances>
+                <Mandatory>Mandatory</Mandatory>
+                <Type>Corelnk</Type>
+                <RangeEnumeration/>
+                <Units/>
+                <Description>One or more CoRE Links to Resource instances from which the values will be stored in the Compact Data array. Each link results in a new item in the Compact Data array: first link points to the resource instance with the value for the first array item, second link to the instance with the second array item, etc. For single-instance resources the Resource instance ID can be omitted from the link, but for multi-instance Resources it MUST be present.</Description>
```
dnav: Suggestion for clarity:

```suggestion
                <Description>One or more CoRE Links to Readable Single Resources and/or Resource Instances of Readable Single Resources, from which the values will be stored in the Compact Data array. Each link results in a new item in the Compact Data array: first link points to the resource instance with the value for the first array item, second link to the instance with the second array item, etc.</Description>
```

dnav (commented):
Hi,

Some comments on the description of the Resources.

Otherwise, there are still some details to be sorted out as discussed.

Regards,

jpradocueva:
@akeranen the latest changes to the object have failed validation. It appears that the original changes have not been mapped to the `history_version` folder.

akeranen:
> @akeranen the latest changes to the object have failed validation. It appears that the original changes have not been mapped to the `history_version` folder.

Indeed, the latest commit by Matt & David is not reflected in the `history_version`. I think we should automate this somehow to avoid the issue in general. Maybe with hard links or post-commit hooks to make sure the versions are in synch? Or if we had only old versions (not the latest being updated) in the `history_version`. Anyway, I'll take a look at this object in general soon.

mkgillmore:
@akeranen   

This PR still has issues with Object Validation tools with error code #1

mkgillmore:
@mkgillmore @jaimejim Will take a look at resolving

mkgillmore (approved):

511.xml
```diff
@@ -0,0 +1,89 @@
...
+        <Resources>
+            <Item ID="0">
+                <Name>Target Links</Name>
+                <Operations>RW</Operations>
+                <MultipleInstances>Single</MultipleInstances>
+                <Mandatory>Mandatory</Mandatory>
+                <Type>Corelnk</Type>
+                <RangeEnumeration/>
+                <Units/>
+                <Description>One or more CoRE Links to Readable Single Resources and/or Resource Instances of Readable Single Resources in the same LwM2M Client, from which the values will be stored in the Compact Data array. Each link corresponds to an item in the Compact Data array: first link points to the resource instance with the value for the first array item, second link to the instance with the second array item, etc.</Description>
```
dnav: To avoid corner cases, I suggest to add a sentence like:

> The links must not point to the Compact Data Resource (ID: 1) of the same Compact Composite Object Instance.

dnav: Could we remove the "in the same LwM2M Client" part ? In the case of a LwM2M Gateway, it could be useful to retrieve values from several IoT Devices.

dnav:
```suggestion
                <Description>One or more CoRE Links to Readable Single Resources and/or Resource Instances of Readable Single Resources in the same LwM2M Client, from which the values will be stored in the Compact Data array. Each link corresponds to an item in the Compact Data array: first link points to the Resource or Resource Instance with value for the first array item, second link to the Resource or Resource Instance with the second array item, etc.</Description>
```

511.xml
```diff
@@ -0,0 +1,89 @@
...
+            </Item>
+            <Item ID="1">
+                <Name>Compact Data</Name>
+                <Operations>R</Operations>
+                <MultipleInstances>Single</MultipleInstances>
+                <Mandatory>Mandatory</Mandatory>
+                <Type>Opaque</Type>
+                <RangeEnumeration/>
+                <Units/>
+                <Description>CBOR array of resource values in the same order as the links in the Target Links resource. CBOR simple value of undefined should be used as the array item if an error occurs while reading the value for the corresponding resource (e.g., link pointing to a resource that doesn't exist or the resource is not allowed to be read because of access control rules).</Description>
```
dnav, line 75:
```suggestion
                <Description>CBOR array of Resource or Resource Instance values in the same order as the links in the Target Links Resource. CBOR simple value of undefined should be used as the array item if an error occurs while reading the value for the corresponding Resource or Resource Instance (e.g., link pointing to a Resource that doesn't exist or the Resource is not allowed to be read because of access control rules).</Description>
```

511.xml
```diff
@@ -0,0 +1,89 @@
...
+POSSIBILITY OF SUCH DAMAGE.
+The above license is used as a license under copyright only.  Please
+reference the OMA IPR Policy for patent licensing terms:
+https://www.omaspecworks.org/about/intellectual-property-rights/
+
+-->
+<LWM2M xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="http://openmobilealliance.org/tech/profiles/LWM2M-v1_1.xsd">
+    <Object ObjectType="MODefinition">
+        <Name>Compact Composite</Name>
+        <Description1>This Object enables collecting a set of Resource values from other Objects into a single resource representation.</Description1>
```
dnav:
```suggestion
        <Description1>This Object enables collecting a set of Resource values from other Objects into a single Resource representation.</Description1>
```

dnav:
Here is an example. I tried to integrate it in the Description2 field but I could not pass the validation because of the formating:

Example using the Example LwM2M Client from the Appendix F of LightweightM2M Technical Specification Core document:

If the Target Links Resource contains the value `"</3/0/0>,</3/0/9>,</1/0/1>"`, a Read operation on the Compact Data Resource performed by the Example LwM2M Server 1 would return the value:
```
83                                      # array(3)
   74                                   # text(20)
      4F70656E204D6F62696C6520416C6C69616E6365 # "Open Mobile Alliance"
   18 64                                # unsigned(100)
   1A 00015180                          # unsigned(86400)
```
which translates in the CBOR diagnostic notation to:
```
["Open Mobile Alliance", 100, 86400]
```

 The same Read operation on the Compact Data Resource performed by the Example LwM2M Server 2, which does not have the Read access right on the Object Instance 0 of the LwM2M Server Object, would return the value:
```
83                                      # array(3)
   74                                   # text(20)
      4F70656E204D6F62696C6520416C6C69616E6365 # "Open Mobile Alliance"
   18 64                                # unsigned(100)
   f7                                   # undefined
```
which translates in the CBOR diagnostic notation to:
```
["Open Mobile Alliance", 100, undefined]
```

akeranen:
Thank you very much for the example @dnav! Examples are great, but I think it's better we give them in the TS doc instead of in the Object spec since the descriptions become easily pretty cluttered and it's going to be hard to provide reasonable formatting for these in the usual places where descriptions are used. But we should definitely use this in the TS!

jpradocueva:
@akeranen are you considering that we should have a TS as part of the release of this Object? The Object was developed directly in the OMA LwM2M Registry because there was no plan to create a TS. If we are planning to create a TS, it should be defined in a private repo as per OMA process.

akeranen:
@jpradocueva I'm thinking we can do a TS later but no need to delay finalising the object publication while waiting for the TS to finish.

jpradocueva:
@akeranen ok, I agree. My point was about developing a TS. It should not be developed in the LwM2M2-Registry repo but in an OMA internal repo.

akeranen:
OK, let's create that repo then and get started. Thanks!

511.xml
kFYatek:
```suggestion
                <Description>One or more CoRE Links to Readable Single Resources and/or Resource Instances of Readable Multiple Resources in the same LwM2M Client, from which the values will be stored in the Compact Data array. Each link corresponds to an item in the Compact Data array: first link points to the resource instance with the value for the first array item, second link to the instance with the second array item, etc.</Description>
```

kFYatek: Additionally, one thing that is IMO under-specified: how much validation should be performed when processing a Write on this resource, vs. only when generating Compact Data? I.e., is referencing an Object that does not exist result in a "Bad Request" response, or should it be accepted, but result in an undefined value in Compact Data?

I personally believe that only the syntactic correctness of the Corelnk value should be validated at Write time, and all other problems result in undefined values in Compact Data - but it wouldn't harm to state this explicitly.

PR #701, opened 2023-02-27
Ucifi/object 6

gmazzoleni:
As uCIFI Alliance we would like to propose to change the Latitude, Longitude and Altitude resources to read and write in order to allow the user to alter the position of a device on field (and store such modification on the device itself).

The change would allow backward compatibility since it would extend the capabilities of the device and trying to alter the position on a device that has not been aligned to the latest data model would simply return "Method not allowed"


gmazzoleni:
The pull request is mostly to allow proposed changes to be reviewed. for such reason it has been marked as draft

dnav:
Hi,

Regarding your use cases:

> allow the user to alter the position of a device on field (and store such modification on the device itself).

Note that the LwM2M Bootstrap-Server can modify or create the values of the Location Object during the Bootstrap sequence using the Bootstrap-Write operation. (We assumed the use case is to set the position of an asset once deployed, and not to indicate to where to go to a motorized device like a drone.)

> In addition, we would also like to to add a few resources to configure a distance threshold over which an alarm may be triggered (by setting the corresponding resource)

A dedicated Object would be better. On the device registration, the LwM2M Server would immediately determine that the device features this geofencing capability. If this is done by adding new Resources to an existing Object, the LwM2M Server would need to perform a Discover or Read operation.

Also note that a simple geofencing is already possible by setting an Observation on the Latitude and Longitude resources, with or without the "Greater Than" and "Less Than" attributes.

Regards,

gmazzoleni:
Thank you very much David for your feedbacks.

I understand your point about using the Bootstrap write, that may also ignore ACL, but still do not completely agree, for a few reasons:
- Bootstrap interface may not be used
- Bootstrap server may not have such information
- Data model should also allow user to alter configuration parameters, eventually if the device thinks this should not be allowed (e.g. because they are being automatically retrieved from other means) would sitll be able to deny the operation

Instead, related to second comment I completely agree and I have already moved such changes to a dedicated object, not to mess the location object. Sorry, I forgot to update the comment accordingly.

Best regards,
Giulio

gmazzoleni:
The functionality will be implemented within uCIFI Data Model.

PR #728, opened 2023-08-24
uCIFI Data Model 2.0: Adding timestamp and measurement quality resources to various objects

gmazzoleni:
**To be reviewed by ignoring whitespace changes (w=1) to the Copyright notice**

Object with minimal changes
	- Version changed from 1.0 to 1.1
	- Addition of timestamp resources
	- Addition of quality of measurement

	3411 - uCIFI Battery
	3413 - uCIFI Generic Actuator
	3419 - Photocell
	3423 - Gas meter
	3424 - Water meter
	3425 - Irrigation valve
	3426 - Water quality sensor
	3427 - Pressure monitoring sensor
	3429 - Tilt sensor
	3431 - Parking Sensor
	3433 - Noise sensor
	3434 - People counter


version_history/3424-1_1.xml
```diff
@@ -0,0 +1,217 @@
...
+				<Name>Blocked meter</Name>
+				<Operations>R</Operations>
+				<MultipleInstances>Single</MultipleInstances>
+				<Mandatory>Optional</Mandatory>
+				<Type>Boolean</Type>
+				<RangeEnumeration></RangeEnumeration>
+				<Units></Units>
+				<Description>Set to True if water meter is blocked.</Description>
+			</Item>
+			<Item ID="13">
```
AlexBHQ, line 164: This is an event, how is it supposed to be reset. Should this go in a log instead?

version_history/3426-1_1.xml
```diff
@@ -0,0 +1,227 @@
...
+  INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+  BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+  LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+  CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+  LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+  ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+  POSSIBILITY OF SUCH DAMAGE.
+
+-->
+<LWM2M xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="http://www.openmobilealliance.org/tech/profiles/LWM2M.xsd">
```
AlexBHQ, line 33: This is a general questions that applies to many objects that contains multiple readings like the following one: ph, chlorine, Redox, dissolved gas, etc. How is the Measurement Quality Indicator suppossed to be handled if some readings are high quality while other are low quality?

version_history/3424-1_1.xml
gmazzoleni: This kind of event should possibly auto-clear when the condition that caused it is solved, e.g. by removing the physical obstruction.
I will expand the comment if you agree.

version_history/3426-1_1.xml
gmazzoleni: Good point.
We were kind of inclined to leave to the implementor the burden to compute an overall measurement quality (e.g. keeping the lowest quality indicator retrieved by the various sensors on board)
Do you have any suggestion?
On how to indicate this concept into the data model (can we alter the description of reusable resources?) or to completely exclude such quality indication..

version_history/3424-1_1.xml
AlexBHQ: Ok, yes the auto-clear makes perfect sense to show the actual state of the water meter. The work order creation part is probably of of scope, and can be handled many way with this information anyway, so not issue here.
Thank you,

version_history/3426-1_1.xml
AlexBHQ: Yes, the best would be to be able to specify the quality for each using a list or another mecanism, but it may be a bit too burdensome for the implementer, and it would be hard specify how to map which QI is for which measurement since some are optionnal. Then again, I am not familiar enough with lwm2m yet. How does it work, if a sensor is at fault? Is the last value reported anyway within the object? Is there guidelines/doc on this that you could point me to? For now, I beleive adding a comment that indicates that "the QI should represent the lowest QI found on any of the measurements", could provide benefits in terms of interoperability. IMHO

version_history/3424-1_1.xml
gmazzoleni: Thanks for asking for clarification anyway. Better to have checked.

PR #729, opened 2023-08-24
uCIFI Data Model 2.0: Adding new objects

gmazzoleni:
**New objects**

Connectivity
	3447 - LPWAN Mesh Connectivity
	3448 - LPWAN Mesh Statistics

Sensors
	3422 - Three-phase electrical meter complement
	3445 - Stream Gauge
	3446 - Rain Gauge
	3449 - Anemometer
	3450 - Wet Bulb Globe Temperature
	3451 - Roll over sensor

Schedules
	3452 - Program Scheduler
	3453 - Calendar Rule
	3454 - Program Function
	3455 - Input Filter
	3456 - Output Filter
	3457 - Program Manager
	3458 - Program Monitor

Device to Device communication
	3459 - Distributed Sensor Group
	3460 - Distributed Sensor Message
	3461 - Sensor Group Device Capabilities

### DIFF
--- a/3422.xml
+++ b/3422.xml
@@ -326,7 +326,7 @@
 				<Operations>R</Operations>
 				<MultipleInstances>Single</MultipleInstances>
 				<Mandatory>Optional</Mandatory>
-				<Type>Float</Type>
+				<Type>Integer</Type>
 				<RangeEnumeration/>
 				<Units/>
 				<Description><![CDATA[Number of power failures in any phase]]></Description>
@@ -336,7 +336,7 @@
 				<Operations>R</Operations>
 				<MultipleInstances>Single</MultipleInstances>
 				<Mandatory>Optional</Mandatory>
-				<Type>Float</Type>
+				<Type>Integer</Type>
 				<RangeEnumeration/>
 				<Units/>
 				<Description><![CDATA[Number of long power failures in any phase.]]></Description>
--- a/3422.xml
+++ b/3422.xml
@@ -138,7 +138,7 @@
 				<Mandatory>Mandatory</Mandatory>
 				<Type>Float</Type>
 				<RangeEnumeration/>
-				<Units>VAR</Units>
+				<Units>var</Units>
 				<Description><![CDATA[System var: Inst. Value. Sum Li Reactive power+ (QI+QII).]]></Description>
 			</Item>
 			<Item ID="11">
@@ -148,7 +148,7 @@
 				<Mandatory>Optional</Mandatory>
 				<Type>Float</Type>
 				<RangeEnumeration/>
-				<Units>VAR</Units>
+				<Units>var</Units>
 				<Description><![CDATA[System var: Inst. Value. Sum Li Reactive power- (QIII+QIV).]]></Description>
 			</Item>
 			<Item ID="12">
--- a/3422.xml
+++ b/3422.xml
@@ -1,0 +1,502 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- BSD-3 Clause License
+
+  Copyright 2023 uCIFI Alliance
+
+  Redistribution and use in source and binary forms, with or without
+  modification, are permitted provided that the following conditions
+  are met:
+
+  1. Redistributions of source code must retain the above copyright
+  notice, this list of conditions and the following disclaimer.
+  2. Redistributions in binary form must reproduce the above copyright
+  notice, this list of conditions and the following disclaimer in the
+  documentation and/or other materials provided with the distribution.
+  3. Neither the name of the copyright holder nor the names of its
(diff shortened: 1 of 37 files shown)


3422.xml
```diff
@@ -0,0 +1,502 @@
...
+				<RangeEnumeration/>
+				<Units>VAh</Units>
+				<Description><![CDATA[Apparent energy export, unified rate, on the 3 lines.]]></Description>
+			</Item>
+			<Item ID="29">
+				<Name>Number of power failures in any phase</Name>
+				<Operations>R</Operations>
+				<MultipleInstances>Single</MultipleInstances>
+				<Mandatory>Optional</Mandatory>
+				<Type>Float</Type>
```
akeranen: Can the number of failures be a non-integer value? Should this be an unsigned integer instead of float?

3452.xml
```diff
@@ -0,0 +1,167 @@
...
+				<Type>Integer</Type>
+				<Operations>RW</Operations>
+				<Mandatory>Mandatory</Mandatory>
+				<MultipleInstances>Single</MultipleInstances>
+				<Description><![CDATA[Number of Program Function object instance containing information on how to control the output(s) based on time and sensor input(s)]]></Description>
+				<Units/>
+			</Item>
+			<Item ID="5">
+				<Name>Input Sources</Name>
+				<Type>String</Type>
```
akeranen, line 98: Would CoRE link make sense here as type?

3459.xml
```diff
@@ -0,0 +1,272 @@
...
+		<MultipleInstances>Multiple</MultipleInstances>
+		<Mandatory>Optional</Mandatory>
+		<Resources>
+			<Item ID="0">
+				<Name>Group State</Name>
+				<Type>Float</Type>
+				<Operations>R</Operations>
+				<Mandatory>Optional</Mandatory>
+				<MultipleInstances>Single</MultipleInstances>
+				<Description><![CDATA[The current output state of the sensor Group, derived either from the latest sensor state published to the group, or computed based on all sensor inputs when data Fusion Operation is used [default=NaN]]]></Description>
```
akeranen, line 50: How is NaN represented on the wire?
(same comment applies to other objects too)

3459.xml
```diff
@@ -0,0 +1,272 @@
...
+				<Type>Time</Type>
+				<Operations>R</Operations>
+				<Mandatory>Optional</Mandatory>
+				<MultipleInstances>Single</MultipleInstances>
+				<Description><![CDATA[SUBSCRIBER ONLY UNIX Timestamp for when the last activity in the group was recorded, either by the node sending a message to the group our retrieved from the "Publishing Timestamp" of the last message received from the group. [default=0]]]></Description>
+				<Units/>
+			</Item>
+			<Item ID="2">
+				<Name>Most recent Sensor Notification</Name>
+				<Type>String</Type>
```
akeranen, line 64: Also here, and other places with "links" [CoRE link](https://www.openmobilealliance.org/release/LightweightM2M/V1_2-20201110-A/HTML-Version/OMA-TS-LightweightM2M_Core-V1_2-20201110-A.html#11-0-Appendix-C-Data-Types-Normative) could make sense. But that would require later LwM2M version than 1.0 IIRC.

3422.xml
gmazzoleni: Thank you Ari for pointing out.
You are perfectly right.
Changing to Integer (in order to keep compatibility with LWM2M 1.0).
Would it be fine?

3452.xml
gmazzoleni: Thank you very much Ari and good point!
That could indeed make sense and could be considered very good practice.

Allow me to check internally to the Alliance if we can consider such change, in order to have a better formal description of the object, despite breaking compatibility with LWM2M 1.0

As an intermediate approach, could also be possible to keep String data type using Core Link syntax.

This also applies to some other resources as well, by the way (as you also commented below).

3459.xml
gmazzoleni: Same as above. Thanks for pointing out

3459.xml
gmazzoleni: Considering that data type is Floatign Point, we were assuming to use NaN floating point encoding.
In content formats such as TLV, JSON (and SenML-JSON) or CBOR (and SenML-CBOR) a standard representation exists.
I don't know about plain text syntax, though would assume "NaN"
Do you suggest any change in the approach?

3452.xml
gmazzoleni: Plan is to keep String data type, but changing the description to mention the usage of Core Link format

gmazzoleni:
Is there anything else needed from uCIFI side related to this pull request?
Otherwise, it could be merged.
Thank you.

PR #762, opened 2024-05-24
[ADMIN] add new entries as reservation placeholders for 7 objects

gocadimic:
> Note:

### DIFF
--- a/DDF.xml
+++ b/DDF.xml
@@ -6191,6 +6191,104 @@ The Object also allows steering the device towards a particular PLMN, or, to a p
         <TSLink>0</TSLink>
     </Item>
     <Item>
+        <ObjectID>10487</ObjectID>
+        <URN>urn:oma:lwm2m:x:10487</URN>
+        <Name>TBD</Name>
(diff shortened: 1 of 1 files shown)


DDF.xml
```diff
@@ -6190,6 +6190,104 @@ The Object also allows steering the device towards a particular PLMN, or, to a p
...
+        <ObjectID>10487</ObjectID>
+        <URN>urn:oma:lwm2m:x:10487</URN>
+        <Name>TBD</Name>
+        <Description>Reservation 10487</Description>
+        <Owner>TBD</Owner>
+        <Source>2</Source>
+        <Ver>1.0</Ver>
+        <DDF></DDF>
+        <Vorto></Vorto>
+        <DDFLink>1</DDFLink>
```
jpradocueva, line 6203: @gocadimic, the DDFLink value must be set to "0" when the DDF file is not present.
I don't think that this is the problem of not rendering the Objects, but let's try to resolve this first in production. We may have a conflict when we try to merge the branch for SEW , 
but we will deal with that problem later.

PR #805, opened 2024-12-23
[ADMIN] add obj 10494 v1.1 reservation placeholder for autosen

gocadimic:
> Note:


jpradocueva (commented):
@gocadimic let's book sometime for us to review this request. The request indicates they will create a new revision of the Object v1.1, this means that we need to prepare a feature-branch for the submitter where we modify the location of the old object and a new entry for the new object as you did, but reflecting a new version. I am closing the PR  as we don't need to reserve an entry on the `prod` branch, the object ID is already taken.

PR #828, opened 2025-05-27
EDHOC Object definition.

dnav:
> Note:

### DIFF
--- a/29.xml
+++ b/29.xml
@@ -60,7 +60,8 @@ LEGAL DISCLAIMER
 <LWM2M xmlns:xsi = "http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="http://www.openmobilealliance.org/tech/profiles/LWM2M-v1_1.xsd" >
     <Object ObjectType="MODefinition">
         <Name>EDHOC</Name>
-        <Description1><![CDATA[This LwM2M Object provides the keying material and configuration to derive a security context using Ephemeral Diffie-Hellman Over COSE (EDHOC) defined in RFC 9528.]]></Description1>
+        <Description1><![CDATA[This LwM2M Object provides the keying material and configuration to derive a security context using Ephemeral Diffie-Hellman Over COSE (EDHOC) defined in RFC 9528. These LwM2M Object Resources MUST only be
+changed by a LwM2M Bootstrap-Server or Bootstrap from Smartcard and MUST NOT be accessible by any other LwM2M Server.]]></Description1>
         <ObjectID>29</ObjectID>
         <ObjectURN>urn:oma:lwm2m:oma:29</ObjectURN>
         <LWM2MVersion>1.1</LWM2MVersion>
--- a/29.xml
+++ b/29.xml
@@ -15,7 +15,7 @@ NORMATIVE INFORMATION
 
   This is available at http://www.openmobilealliance.org/
 
-  Send comments to technical-comments@mail.openmobilealliance.org
+  Send comments to helpdesk@omaorg.org
 
 CHANGE HISTORY
 
--- a/29.xml
+++ b/29.xml
@@ -1,0 +1,166 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<!--
+FILE INFORMATION
+
+OMA Permanent Document
+   File: 
+   Type: xml
+
+Public Reachable Information
+   Path: http://www.openmobilealliance.org/tech/profiles
+   Name: 
+
+NORMATIVE INFORMATION
+
+  Information about this file can be found in the latest revision of
+
+    
+
+  This is available at http://www.openmobilealliance.org/
+
+  Send comments to technical-comments@mail.openmobilealliance.org
+
+CHANGE HISTORY
+
+    
+LEGAL DISCLAIMER
+
+-->
+
+<LWM2M xmlns:xsi = "http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="http://www.openmobilealliance.org/tech/profiles/LWM2M-v1_1.xsd" >
+    <Object ObjectType="MODefinition">
+        <Name>EDHOC</Name>
+        <Description1><![CDATA[This LwM2M Object provides the keying material and configuration to derive a security context using Ephemeral Diffie-Hellman Over COSE (EDHOC) defined in RFC 9528.]]></Description1>
+        <ObjectID>29</ObjectID>
+        <ObjectURN>urn:oma:lwm2m:oma:29</ObjectURN>
+        <LWM2MVersion>1.1</LWM2MVersion>
+        <ObjectVersion>1.0</ObjectVersion>
+        <MultipleInstances>Multiple</MultipleInstances>
+        <Mandatory>Optional</Mandatory>
+        <Resources>
+            <Item ID="0">
+                <Name>Peer Public Key Type</Name>
+                <Operations></Operations>
+                <MultipleInstances>Single</MultipleInstances>
+                <Mandatory>Mandatory</Mandatory>
+                <Type>Integer</Type>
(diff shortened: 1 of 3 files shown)


mkgillmore (commented):
Needs proper license for OMA

29.xml
```diff
@@ -0,0 +1,181 @@
...
+
+NORMATIVE INFORMATION
+
+  Information about this file can be found in the latest revision of
+
+    
+
+  This is available at http://www.openmobilealliance.org/
+
+  Send comments to technical-comments@mail.openmobilealliance.org
```
jpradocueva:
```suggestion
  Send comments to helpdesk@omaorg.org
```

version_history/29-1_0.xml
```diff
@@ -0,0 +1,181 @@
...
+
+NORMATIVE INFORMATION
+
+  Information about this file can be found in the latest revision of
+
+    
+
+  This is available at http://www.openmobilealliance.org/
+
+  Send comments to technical-comments@mail.openmobilealliance.org
```
jpradocueva:
```suggestion
  Send comments to helpdesk@omaorg.org
```

jpradocueva (approved):
Update to email address.

PiotrRoszkowski:
Hi David,

First off, thanks for putting together the EDHOC object definition. It’s a solid proposal and, in my opinion, definitely something that should be added to LwM2M. As I was going through it, a few questions and potential improvements came to my mind. I’ve listed them below, hope you find them useful. Happy to discuss further if anything’s unclear :)

1. Is this object also intended to establish the OSCORE Security Context for connections with the LwM2M Bootstrap Server? If so, we should clarify that only one instance of this object may correspond to the Bootstrap Server. This constraint should be explicitly stated.
2. There are two scenarios according to Resource 9. When it is set to True, the purpose of the EDHOC object is strictly to establish an OSCORE Security Context. If that's the case, shouldn’t we clarify how this EDHOC object instance links to a specific OSCORE object instance? Especially considering that the OSCORE instance is tied to a specific LwM2M Security object instance? How do we define this relationship? The second option is with Resource 9 set to False. Then the EDHOC object instance is not meant for OSCORE. What then? What are the intended use cases? Should we specify them somehow? Or maybe we can just disallow such a scenario for now and consider adding support for it in a future version of the object, if needed?
3. I believe this object should only be modified by the LwM2M Bootstrap Server or via Bootstrap from Smartcard. This restriction should be explicitly stated, just as it is for the LwM2M Security Object and the LwM2M OSCORE Object.
4. Moreover, it seems that there are at least a few places in the Core and Transport specifications that should be updated to reflect the introduction of this object. A few examples come to my mind:
- **Core spec, Section 6.1.7.6 Bootstrap-Delete Operation** currently states:
  > The two exceptions are the LwM2M Bootstrap-Server Account, potentially including an associated Instance of an OSCORE Object (ID:21), and the single Instance of the mandatory Device Object (ID:3), which are not affected by any Delete operation.

  Shouldn't this list also include an associated EDHOC Object Instance? Other similar sections are in the Core spec: 6.1.7.3 — Bootstrap-Discover Operation, 6.1.7.7 — Bootstrap-Pack-Request Operation, 6.2.1 — Register Operation

- **Transport spec, Section 5. Security** currently states:
  > The LwM2M specification also supports application layer security based on OSCORE with pre-shared secrets."

  If I understand correctly, with the introduction of EDHOC, we can remove "pre-shared secrets" part of the sentence?

- Finally, **Transport spec, Section 5. Security** - shouldn't this section explicitly mention EDHOC or even include a dedicated subsection about it?

Looking forward to your thoughts!

29.xml
```diff
@@ -0,0 +1,181 @@
...
+            <Item ID="4">
+                <Name>Private Key</Name>
+                <Operations></Operations>
+                <MultipleInstances>Single</MultipleInstances>
+                <Mandatory>Optional</Mandatory>
+                <Type>Opaque</Type>
+                <RangeEnumeration></RangeEnumeration>
+                <Units></Units>
+                <Description><![CDATA[Stores the LwM2M Client's private key.
+If this Resource is null or not instantiated, the LwM2M Client is to use OOB credentials (e.g. by using EST), and the Client Public Key Resource must be null or not instantiated.]]></Description>
```
eliasweingaertner, line 122: Saying that the "resource is null" is vague: What does the private key need to contain? Does it mean that an internal pointer in the implementation points to NULL? Or does it mean it contains a single, null byte? ?

29.xml
```diff
@@ -0,0 +1,181 @@
...
+        <ObjectID>29</ObjectID>
+        <ObjectURN>urn:oma:lwm2m:oma:29</ObjectURN>
+        <LWM2MVersion>1.1</LWM2MVersion>
+        <ObjectVersion>1.0</ObjectVersion>
+        <MultipleInstances>Multiple</MultipleInstances>
+        <Mandatory>Optional</Mandatory>
+        <Resources>
+            <Item ID="0">
+                <Name>Peer Public Key Identifier</Name>
+                <Operations></Operations>
```
eliasweingaertner, line 74: Can you explain why none of these resources carry an operation? Without an operation, the information is essentially hidden on the device; it can't be retrieved, written or executed using standard writes, reads or executes. 

Maybe I don't understand the purpose of the object fully, but what is the reason to specify an object if none of the information is accessible?

PiotrRoszkowski: I believe it results from restriction that this object should be accessed only by LwM2M Bootstrap Server. The same lack of `Operations` approach is used for LwM2M Security, LwM2M OSCORE and LwM2M COSE objects:)

eliasweingaertner: > I believe it results from restriction that this object should be accessed only by LwM2M Bootstrap Server. The same lack of `Operations` approach is used for LwM2M Security, LwM2M OSCORE and LwM2M COSE objects:)

That was my assumption also, hence I asked for clarification. But for example Public Keys technically could be readable...

mkgillmore:
@dnav Please have a look at the comments submitted.  I would like to resolved on 6/24 if possible

29.xml
dnav: I assumed it means a zero-length buffer.
I followed the wording used in the LwM2M Security Object and the Firmware Update Object in the Core spec. But after re-reading these sections, I think we can improve the Core spec by providing a clear definition of a null Opaque or String resource.

29.xml
dnav: In this, I followed the convention used in the Objects mentionned by @PiotrRoszkowski. This Object is only accessible to the LwM2M Bootstrap Server.

dnav:
Replying to @PiotrRoszkowski comment:

Our plan here has several steps. The first one is to specify the EDHOC Object which can be used in various scenarii.

The next steps are to make use of the EDHOC Objects in the next LwM2M specification to secure the connection between the LwM2M Clients and Servers.  At this stage, I hope we will address all your points.

Best regards

mkgillmore:
@dnav any updates on this?

mkgillmore:
Move out of core objects and publish in the OMA object range. @mkgillmore will make the change

mkgillmore:
Object 514 has has been allocated

marco-tiloca-sics:
After having re-checked the PR, there seems to be an inconsistency in the definition of the EDHOC Object that you may want to fix.

Unlike the analogous Object Resource "Peer Public Key Identifier" (ID=0) that has type opaque, the Object Resource "Client Key Identifier" (ID=2) has type integer. Its description says:

> The ID_CRED_x of the credential stored in Client Public Key. If this Resource is null or not instantiated, the LwM2M Client is to generate the ID_CRED_x matching the Client Public Key.

This is maybe an oversight and the correct type of "Client Key Identifier" should be opaque (just like for the Object Resource "Peer Public Key Identifier" specifying the ID_CRED_X of the LwM2M Server).

PR #837, opened 2025-06-25
XML to JSON Complete

Tpate33:
> Note:


jpradocueva:
@mkgillmore If this PR is merged, our validation, editor, and publication apps will not be able to validate, edit or publish these JSON objects. These systems were developed for .xml files. It can be updated, but we should prepare a plan.

jpradocueva:
@mkgillmore, also, the version_history folder is not available. This PR will only display the latest versions of the objects.

jpradocueva:
@mkgillmore, our validation system does **not** validate the JSON files directly. Since there were no changes to the `.xml` files, the system marked the submission as Passed Validation. However, this status only means that no issues were found in the .xml files—it does not imply that the JSON objects were validated or are free of errors.

PR #839, opened 2025-07-07
Network routing tree - first dev

LMPPTSCH:
> Note:


gmazzoleni:
@LMPPTSCH Isn't the information contained here a duplicate of what is already present in object 3447?
Node address and parent address?

LMPPTSCH:
> @LMPPTSCH Isn't the information contained here a duplicate of what is already present in object 3447? Node address and parent address?

@gmazzoleni, indeed, we may have here an overlap. Check the resources that we aim to use:
    RES_NET_ROUTE_TREE_NODE_ID = ? (Any suggestions for the node ID?
    RES_NET_ROUTE_TREE_PARENT_ID = 23 (Could you confirm if we can use this one?)
    RES_NET_ROUTE_TREE_TIMESTAMP = 5518 (Could you confirm if we can use this one?)

LMPPTSCH:
We'll use the existing object 3447

PR #855, opened 2025-11-17
Update Common.xml

will-aos:
Add reusable AppID resource. The AppID index corresponds to an application registered in a TBD external Registry (LwM2M or GEISA).

> Note:

### DIFF
--- a/Common.xml
+++ b/Common.xml
@@ -276,6 +276,17 @@ If the LwM2M client is connected to multiple servers, this resource SHOULD be ha
         <TS></TS>       
         <TSLink>0</TSLink>
        </Item>
+       <Item ID="4050">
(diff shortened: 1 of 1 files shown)


mkgillmore (approved):

lodup29:
For clarification: Is the ‘AppID’ intended to be used across different objects (for example, the same as the ‘App ID’ in Object 19, currently resource ID 5)?

will-aos:
> For clarification: Is the ‘AppID’ intended to be used across different objects (for example, the same as the ‘App ID’ in Object 19, currently resource ID 5)?

@lodup29 yes, the intent is to add a resource called AppID that is reusable across multiple objects

PR #881, opened 2026-03-01
Geisa obj 3601 placeholder

will-aos:
> Note:


gocadimic:
@will-aos,
This PR doesn't pass validation due to the units. 
In object 3601, the unit value "MB" doesn't exist in the Senml registry used by OMA LwM2M for units; see https://www.iana.org/assignments/senml/senml.xhtml.
The alternative is to use "KiB" (kibibyte) or GB (gigabyte).
Please evaluate which unit you want to apply; whichever unit you choose must also be applied to the version_history. We can apply these changes for you if you let us know the unit.

jpradocueva:
@will-aos, @mkgillmore, @gocadimic. 

F.Y.I., OMA has an additional internal SenML list of units that can be used during Object validation. It was added to support registrations that use units pending registration with SenML. The file is in this repo, it is called [LWM2M_senml_units.xml](https://github.com/OpenMobileAlliance/lwm2m-registry/blob/prod/LWM2M_senml_units.xml). If new "units" are added to this list, it can be used during the validation by clicking the box of `Allow additional units`. See screenshot below. 

image

PR #929, opened 2026-07-30
Split and refined aggregate and component platform monitoring

swegener-rtp:
## Summary

Refines the GEISA platform monitoring objects so that:

* Object 3605 provides mandatory, single-instance aggregate platform monitoring
* Object 3608 provides (currently) optional, multiple-instance logical component monitoring
* Aggregate monitoring remains useful without requiring enumerating of component instances
* Component-level detail remains available for targeted investigations if/as required

## Object 3605

`PlatformMonitoring` now provides the authoritative aggregate platform monitoring view.

This update:

* keeps aggregate platform identity, state, registration, communication, queue, API, event, and monitoring telemetry;
* renames the primary state to `Platform Monitoring State`;
* replaces the free-form health reason with a machine-readable `Monitoring Reason Code`;
* adds optional human-readable `Reason Detail`;
* adds optional multi-instance `Related Component` links to relevant Object 3608 instances;
* removes subsystem-specific queue, API broker, message bridge, event log, and local monitoring state resources;
* retains aggregate counters and timestamps without renumbering the remaining resources.

## Object 3608

`PlatformComponentMonitoring` provides detailed monitoring for logical platform components, services, functions, queues, APIs, and operational paths.

This update:

* renames the primary state to `Component Monitoring State`;
* uses the same monitoring-state values as Object 3605;
* uses the same machine-readable `Monitoring Reason Code` values;
* adds optional human-readable `Reason Detail`;
* uses `Component Role` with the generic component monitoring state instead of separate role-specific health-state resources;
* retains role-relevant counters, timestamps, registration, communication, queue, API, event, and execute resources.

## Shared monitoring model

The shared monitoring states are:

* Unknown
* Healthy
* Degraded
* Failed
* Disabled
* Not reporting
* Starting
* Stopping
* Stopped
* Blocked by policy

The shared monitoring reason codes provide machine-readable causes such as component failure, stale monitoring data, communication failure, registration failure, queue pressure, resource pressure, authorization denial, security restriction, internal error, and timeout.

`Reason Detail` provides optional human-readable diagnostic context and is required when the reason code is `Other`. Consumers must not parse it for interoperable behavior.

## Aggregate and component behavior

Object 3605 remains the authoritative aggregate view. An EMS can observe or read it without enumerating all Object 3608 instances.

Object 3608 provides component-level detail when investigation is needed. Object 3605 may link directly to one or more related Object 3608 instances, but remains meaningful when Object 3608 is not implemented or no component link is available.

## Validation

Verified:

* all four XML files validate against `LWM2M-v1_1.xsd`;
* root and version-history copies are byte-identical;
* resource IDs are unique and ordered;
* removed resource IDs remain unused;
* `git diff --check` passes;
* the change is limited to:

  * `3605.xml`
  * `3608.xml`
  * `version_history/3605-1_0.xml`
  * `version_history/3608-1_0.xml`


jpradocueva:
@swegener-rtp, @mkgillmore.
In this PR, I applied the following changes:
* Object 20, version 3.1 was already agreed and published by OMA DMSO. I removed it from the PR
* Object 3, versioin 1.3 was already agreed and published by OMA DSMO. I remove it from the PR
* Object 32666, I also removed it from the PR
* Object 514: I removed it from the PR, as it is already in production brunch
* Object 607: I removed it from the PR, as this object is already in the production brunch. 
* Object 9, version 1.1. I removed it from the PR as it is also in the production brunch.

swegener-rtp:
Thanks @jpradocueva . This should be good to merge to the branch on this one as is now. CC @mkgillmore

PR #936, opened 2026-09-10
[SCWG] Fixed incorrect XML format of object 3457

gmazzoleni:
It has been found that XML format for object 3457 in version 1.1 is incorrect due to early closure of Resources tag.
Proposal is to fix and replace existing XML version without incrementing any version, since no semantic change is being indeed introduced.


mkgillmore:
@jpradocueva Will review and merge if there are no issues found

jpradocueva:
Object 3457 stored in the `prod` branch failed validation. For some reason, the Validation tool isn't detecting that the Object is not well-formed and valid. 
The group agreed to merge this PR as soon as possible, as it is incorrect in production.

